Issue #208, opened 2014-07-29
validate namespace prefix in nodes before doing lookups (and writes?)


lsmith77:
Jackrabbit throws a NamespaceException when attempting to read a node with a node name that includes a namespace prefix that has not been registered.

We should also add a PHPCR API test for this.

dbu:
is it essential to have this in the 1.2 milestone?

dbu:
#223 handles the writing case as the NodeProcessor validates the namespace.

for reading i just created #244

PR #223, opened 2014-10-18
Updated to use the new node processor in jackalope

dantleech:
Updated to use the new (code merge pending) node processor in jackalope:
https://github.com/jackalope/jackalope/pull/245

fixes part of #208

needs https://github.com/jackalope/jackalope/pull/279

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -4,10 +4,8 @@ namespace Jackalope\Transport\DoctrineDBAL;
 
 use PHPCR\LoginException;
 use PHPCR\NodeType\NodeDefinitionInterface;
-use PHPCR\NodeType\NodeTypeDefinitionInterface;
 use PHPCR\NodeType\NodeTypeExistsException;
 use PHPCR\NodeType\PropertyDefinitionInterface;
-use PHPCR\PropertyInterface;
 use PHPCR\RepositoryInterface;
 use PHPCR\NamespaceRegistryInterface;
 use PHPCR\CredentialsInterface;
@@ -22,7 +20,6 @@ use PHPCR\ItemExistsException;
 use PHPCR\ItemNotFoundException;
 use PHPCR\ReferentialIntegrityException;
 use PHPCR\SimpleCredentials;
-use PHPCR\ValueFormatException;
 use PHPCR\PathNotFoundException;
 use PHPCR\Query\InvalidQueryException;
 use PHPCR\NodeType\ConstraintViolationException;
@@ -41,7 +38,6 @@ use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Jackalope\Node;
 use Jackalope\Property;
 use Jackalope\Query\Query;
-use Jackalope\Transport\AddNodeOperation;
 use Jackalope\Transport\MoveNodeOperation;
 use Jackalope\Transport\BaseTransport;
 use Jackalope\Transport\QueryInterface as QueryTransport;
@@ -52,7 +48,6 @@ use Jackalope\Transport\TransactionInterface;
 use Jackalope\Transport\StandardNodeTypes;
 use Jackalope\Transport\DoctrineDBAL\Query\QOMWalker;
 use Jackalope\NodeType\NodeTypeManager;
-use Jackalope\NodeType\NodeType;
 use Jackalope\NodeType\NodeTypeDefinition;
 use Jackalope\FactoryInterface;
 use Jackalope\NotImplementedException;
@@ -140,9 +135,9 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     protected $namespaces;
 
     /**
-     * @var string|null
+     * @var array The namespaces at initial state, in case of rollback.
      */
-    private $sequenceWorkspaceName;
+    private $originalNamespaces = array();
 
     /**
      * @var string|null
@@ -173,11 +168,6 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     private $referencesToDelete = array();
 
     /**
-     * @var array
-     */
-    private $additionalNodeAddOperations = array();
-
-    /**
      * @var boolean
      */
     private $connectionInitialized = false;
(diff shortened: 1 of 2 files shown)


lsmith77:
needs a rebase

dbu:
rebased. but the tests still fail. there seem to be some glitches and something seems to try to store SimpleCredentials instead of nodes and properties... @dantleech can it be that the NodeProcessor is broken? do you have time to look into this one? its the last major stumbling block for 1.2

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -190,6 +196,18 @@ public function __construct(FactoryInterface $factory, Connection $conn)
...
         $this->conn = $conn;
+        $this->namespaces = new \ArrayObject();
+
+        if ($conn->getDatabasePlatform() instanceof PostgreSqlPlatform) {
+            $this->sequenceWorkspaceName = 'phpcr_workspaces_id_seq';
+            $this->sequenceNodeName = 'phpcr_nodes_id_seq';
+            $this->sequenceTypeName = 'phpcr_type_nodes_node_type_id_seq';
+        }
+
+        // @TODO: move to "SqlitePlatform" and rename to "registerExtraFunctions"?
```
dbu: this still says todo - do we want to do this before merging?

dantleech:
Its quite a risky change for not much gain probably.. I implemented this for the benefit of Jackalope Filesystem, but I don't think it fixes any existing "problems" (?). So I think this could be used in Jackalope 2.0.

Do you think its worth putting in the next release?

dbu:
it would clean up some code and we would avoid that we might still fix issues or add things in validation in jackalope-doctrine-dbal that we then need to synchronize back to base jackalope.

but i do see the risk as well. so maybe keep this for 2.0?

@lsmith77 what do you think?

src/Jackalope/Transport/DoctrineDBAL/Client.php
lsmith77: this is independent of this PR and would require adding this to Doctrine DBAL

lsmith77:
well I think it makes sense to resolve this .. after all the issue could also be in the code that is now merged into Jackalope itself.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: ah i see. sounds like a good idea to start that PR, but i guess we keep that for a while until we can bump the dbal version to only those having the feature.

dbu:
i just rebased this on master. which gives a +1 for doing it for this release. there where conflicts for the validate methods of the Client. which probably means we have to port some fixes over to the NodeProcessor. i have not figured out what was different, as the methods are rather huge and git considered the whole thing different

dbu:
the test failures seem to be about a bug in the NodeProcessor it self, it someplaces confuses arguments and tries to put simple credentials into a property it seems...

lsmith77:
seems to be failing :(

dbu:
yes, i did not fix anything, only rebased, noted the problem and started to look at the test failure output. somebody needs to dig through the node processor.

dbu:
i am on it now, will do a PR against jackalope soon

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -134,9 +135,9 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface,
     private $checkLoginOnServer = true;
 
     /**
-     * @var array
+     * @var \ArrayObject
      */
-    protected $namespaces = array();
+    protected $namespaces;
```
dbu: it would be nice to use the NamespaceRegistry here - but we don't have these things available at the transport level.

composer.json
```diff
@@ -20,7 +20,7 @@
         "doctrine/dbal": "~2.5,>=2.5.0",
         "phpcr/phpcr": "~2.1.2",
         "phpcr/phpcr-utils": "~1.2,>=1.2.4",
-        "jackalope/jackalope": "~1.2.0"
+        "jackalope/jackalope": "dev-fix-nodeprocessor as 1.2.0"
```
dbu: don't forget to revert this before merging

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1767,136 +1785,25 @@ public function reorderChildren(Node $node)
...
-
-                }
-            }
-        }
-
-        foreach ($def->getDeclaredSupertypes() as $superType) {
-            $this->validateNodeWithType($node, $superType);
-        }
+        $this->nodeProcessor = new NodeProcessor(
+            $this->credentials->getUserID(),
```
dbu: this was the reason for most of the test failures - we need the id, not the credentials object.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2517,7 +2363,10 @@ public function rollbackTransaction()
 
         try {
             $this->inTransaction = false;
-            $this->namespaces = array();
+
+            // reset namespaces
+            $this->namespaces->exchangeArray(array());
+            $this->getNamespacesObject();
```
dbu: the problem with the reference is that using the ArrayObject does not trigger a reload. but i guess we need to go with this for now and hope to find a solution to pass the NamespaceRegistry in 2.0

dbu:
hm, that travis output... looks like "something" with rollback breaks and there is a process isolation issue, failing all tests after that one. i can't really see the relation to what i am doing here though. ideas?

dbu:
restarted https://travis-ci.org/jackalope/jackalope-doctrine-dbal/jobs/54830887 which was green (last master build with postgres) to see if travis updated postgres or something like that.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dantleech: @dbu this code is already in master, it was moved to `initConnection`, so I guess this needs a rebase or something.

dbu:
thanks dan. removed the duplicated code - hopefully that was causing the problems with postgres...

dbu:
hm, still the same error on travis. could resetting the namespaces in the middle of the rollback be the culprit?

dantleech:
Just tested locally, and it works in isolation ..

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: this line is messing up the doctrine connection with postgres. after testAddNodeInParallel the connection is already in rollback only mode... i will just copy the original namespaces for rollback. will update the PR once it runs through locally.

dbu:
okay, i think i got the culprit and have a fix. lets see if travis agrees.

dantleech:
all green, good work :)

lsmith77:
like I said in the dependent PR, I do not know the code change in detail but I think it makes sense to "risk" this refactoring in order to improve code reuse.

dbu:
merged the PR on jackalope/jackalope, tagged RC1 and updated composer.json

Issue #228, opened 2014-12-12
`doctrine:phpcr:init:dbal` should have a `--force` flag to prevent DDL statements to be directly executed


Ocramius:
As per reference, see the warning below:

![screenshot](http://marco-pivetta.com/screen/caps/f82edc.png)

dbu:
fixed in #232

PR #232, opened 2014-12-13
added --force option to init command

lsmith77:
fix for https://github.com/jackalope/jackalope-doctrine-dbal/issues/228

### DIFF
--- a/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
+++ b/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
@@ -21,6 +21,10 @@ use Jackalope\Transport\DoctrineDBAL\RepositorySchema;
  */
 class InitDoctrineDbalCommand extends Command
 {
+    const RETURN_CODE_NOT_DROP = 1;
+
+    const RETURN_CODE_NO_FORCE = 2;
+
     /**
      * @see Command
      */
@@ -30,6 +34,10 @@ class InitDoctrineDbalCommand extends Command
             ->setName('jackalope:init:dbal')
             ->setDescription('Prepare the database for Jackalope Doctrine-Dbal.')
             ->setDefinition(array(
+                new InputOption(
+                    'force', null, InputOption::VALUE_NONE,
+                    'Set this parameter to execute this action'
+                ),
                 new InputOption(
                     'dump-sql', null, InputOption::VALUE_NONE,
                     'Instead of try to apply generated SQLs to the database, output them.'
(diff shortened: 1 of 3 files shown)


dbu:
@lsmith77 do you want to try with the mock class for the pdo exception?

lsmith77:
will give it another try today

lsmith77:
@dbu: can you review?

dbu:
thanks

src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
```diff
@@ -58,22 +67,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
             if ($input->getOption('drop')) {
                 foreach ($schema->toDropSql($connection->getDatabasePlatform()) as $sql) {
                     if (true === $input->getOption('dump-sql')) {
                         $output->writeln($sql);
                     } else {
-                        try {
-                            $connection->exec($sql);
-                        } catch (\Exception $e) {
-                            $output->writeln($e->getMessage());
-                        }
```
WouterJ: @lsmith77 do you maybe remember why you removed this try...catch block when adding `--force`?

lsmith77: hmm not sure ... I guess we wanted to give more context in the error output than we get from just outputting the error message.

WouterJ: Hmm, then we should improve the code in the catch block a bit. We still need to catch them, as the command shouldn't fail when there are no databases to drop.

lsmith77: yeah .. or maybe we should have a switch to ignore drop errors ? in theory we could try to look at the error sql code to determine why the query failed, ie if it was because the table didn't exist or because of some other issue.

WouterJ: It's a bit harder than that, as not all backends use the same error codes IIRC. This has been fixed in dbal 2.5, by throwing very specific exceptions. This way, we can catch only `TableNotFoundException`. However, we also need to support dbal 2.4, which throws the generic `DBALException`. Maybe we can do some string matching?

The current behaviour on jackalope-doctrine-dbal master is: Fail if table to drop did not exists, unless `--force` was passed to the command. I think that's the best option to apply to 1.1 too

lsmith77: then maybe lets catch the TableNotFoundException and just output a small hint, we can also catch the DBALException and rethrow it if we are on 2.5+ ?

PR #235, opened 2014-12-18
Test lowest versions of dependencies

WouterJ:
See also https://github.com/jackalope/jackalope-jackrabbit/pull/105


dbu:
see #247, thanks for having started this, wouter!

PR #238, opened 2015-01-05
fix issue with postgres 9

### LINKED ISSUE #230: fix travis setup with PostgreSQL
https://travis-ci.org/jackalope/jackalope-doctrine-dbal/builds/43917369

dbu:
fix #230

was able to reproduce locally with psql 9.3.5. travis does not tell its postgres version in the build system info, only all other information you could think of, so not sure if its psql 9 only.

@lsmith77 can you check this branch on your system to see if the changes work on your postgres version?

i don't understand exactly what is happening or if its intended, but looks like an issue with auto-detecting the type when the value is 0. if this really fixes the issue, we need it anyways as postgres versions with this issue are out there.

@dantleech while debugging, i tried to query the length value in the select. but this query was considered invalid:
```
            SELECT LENGTH(data.[empty-value]) AS length
            FROM [nt:unstructured] AS data
            WHERE
              data.[empty-value] IS NOT NULL
```
is this a known issue, or not supposed to work anyways or should we create an issue for the query parser?

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -840,7 +840,7 @@ class QOMWalker
             return sprintf("EXTRACTVALUE(%s.props, '//sv:property[@sv:name=\"%s\"]/sv:value[%d]/@%s')", $alias, $property, $valueIndex, $attribute);
         }
         if ($this->platform instanceof PostgreSqlPlatform) {
-            return sprintf("(xpath('//sv:property[@sv:name=\"%s\"]/sv:value[%d]/@%s', CAST(%s.props AS xml), %s))[1]::text", $property, $valueIndex, $attribute, $alias, $this->sqlXpathPostgreSQLNamespaces());
+            return sprintf("CAST((xpath('//sv:property[@sv:name=\"%s\"]/sv:value[%d]/@%s', CAST(%s.props AS xml), %s))[1]::text AS bigint)", $property, $valueIndex, $attribute, $alias, $this->sqlXpathPostgreSQLNamespaces());
         }
         if ($this->platform instanceof SqlitePlatform) {
             return sprintf("EXTRACTVALUE(%s.props, '//sv:property[@sv:name=\"%s\"]/sv:value[%d]/@%s')", $alias, $property, $valueIndex, $attribute);


lsmith77:
awesome!

seems like this is an issue with newer versions of PostgreSQL since this code was introduced in 2013 and so far hasn't caused issues. then again maybe we have added some new test code?

https://github.com/jackalope/jackalope-doctrine-dbal/commit/1cb544eb9236755f0b3cc8a506b489f67357380c#diff-92e5f5b1040de502d9db5f5f3f08682aR802

will try to check if things work for me in this branch with my older version of PostgreSQL but I would assume so. the change looks quite safe to me.

dantleech:
@dbu its not valid according to the QOM Length only applies to dynamic operands https://github.com/phpcr/phpcr/blob/master/src/PHPCR/Query/QOM/LengthInterface.php

On 5 January 2015 19:11:41 GMT+00:00, David Buchmann <notifications@github.com> wrote:
>fix #230
>
>was able to reproduce locally with psql 9.3.5. travis does not tell its
>postgres version in the build system info, only all other information
>you could think of, so not sure if its psql 9 only.
>
>@lsmith77 can you check this branch on your system to see if the
>changes work on your postgres version?
>
>i don't understand exactly what is happening or if its intended, but
>looks like an issue with auto-detecting the type when the value is 0.
>if this really fixes the issue, we need it anyways as postgres versions
>with this issue are out there.
>
>@dantleech while debugging, i tried to query the length value in the
>select. but this query was considered invalid:
>```
>            SELECT LENGTH(data.[empty-value]) AS length
>            FROM [nt:unstructured] AS data
>            WHERE
>              data.[empty-value] IS NOT NULL
>```
>is this a known issue, or not supposed to work anyways or should we
>create an issue for the query parser?
>You can merge this Pull Request by running:
>
>git pull https://github.com/jackalope/jackalope-doctrine-dbal
>fix-postgres-query
>
>Or you can view, comment on it, or merge it online at:
>
>  https://github.com/jackalope/jackalope-doctrine-dbal/pull/238
>
>-- Commit Summary --
>
>  * fix issue with postgres 9
>
>-- File Changes --
>
>    M src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php (2)
>
>-- Patch Links --
>
>https://github.com/jackalope/jackalope-doctrine-dbal/pull/238.patch
>https://github.com/jackalope/jackalope-doctrine-dbal/pull/238.diff
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jackalope/jackalope-doctrine-dbal/pull/238

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dbu:
yay!

@dantleech okay. well, SQL2 is not the main thing of jcr. would have seemed a use case, like collecting the length info of binaries to show a list of files without loading the nodes. at least we can load the nodes without the actual binary data...

PR #241, opened 2015-01-18
Ignore drop exception if --force is used

dantleech:
Allow `--force` to ignore exceptions when `--drop` is used.

Does that make sense?

Fixes the current issue with build failures with the CMF testing component.


dantleech:
/cc @lsmith77 

I am not sure why this suddendly stopped working for us. The `--force` PR was merged, but after adding `--force` to the testing travis script, we still have failures relating to `--drop`.

https://travis-ci.org/symfony-cmf/RoutingBundle/jobs/47414182

dbu:
does that help with https://github.com/symfony-cmf/RoutingBundle/pull/281 ?

Issue #239, opened 2015-01-05
race condition with sqlite?


dbu:
sqlite builds fail sometimes. in these 3 builds, each has one fail with a different php version:
https://travis-ci.org/jackalope/jackalope-doctrine-dbal/builds/45984013
https://travis-ci.org/jackalope/jackalope-doctrine-dbal/builds/45983999
https://travis-ci.org/jackalope/jackalope-doctrine-dbal/builds/45984013

can we synchronize / flush something, or do we need to wait at some step to reduce the risk of races?

dantleech:
Since when? I have seen this before, and I think there are some general issues with session state.. but I wonder which commit introduced / revealed this.

dbu:
wonder if we could just remove the .sqlite file and rebuild the database instead. that might even be faster than the remove statements.

dantleech:
or better yet -- keep a copy of the initial file, and copy it everytime. like is done in the LiipFunctionalTest bundle.

PR #243, opened 2015-02-03
[POC] Support for comparing multivalue numeric fields

dantleech:
This is just a POC for now.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -627,10 +627,22 @@ class QOMWalker
         $alias = $this->getTableAlias($propertyOperand->getSelectorName() . '.' . $propertyOperand->getPropertyName());
         $property = $propertyOperand->getPropertyName();
 
-        return
-            $this->sqlXpathExtractNumValue($alias, $property) . " " .
-            $operator . " " .
-            $literalOperand->getLiteralValue();
+
+        if ($this->platform instanceof MySqlPlatform && '=' === $operator) {
+            return sprintf(
+                '0 != FIND_IN_SET("%s", REPLACE(EXTRACTVALUE(%s.props, \'//sv:property[@sv:name="%s"]/sv:value\'), " ", ","))',
+                $literalOperand->getLiteralValue(),
+                $alias,
+                $property
+            );
+        }
+
+        return sprintf(
+            '%s %s %s',
+            $this->sqlXpathExtractNumValue($alias, $property),
+            $operator,
+            $literalOperand->getLiteralValue()
+        );
     }
 
     /**
@@ -831,7 +843,8 @@ class QOMWalker
             return "(xpath('//sv:property[@sv:name=\"" . $property . "\"]/sv:value[1]/text()', CAST($alias.props AS xml), ".$this->sqlXpathPostgreSQLNamespaces()."))[1]::text::int";
         }
 
-        return $this->sqlXpathExtractValue($alias, $property);
+
+        return 'CAST(' . $this->sqlXpathExtractValue($alias, $property) . ' AS DECIMAL)';
     }
 
     private function sqlXpathExtractValueAttribute($alias, $property, $attribute, $valueIndex = 1)


src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -627,10 +627,26 @@ public function walkNumComparisonConstraint(QOM\PropertyValueInterface $property
...
         $property = $propertyOperand->getPropertyName();
 
-        return
-            $this->sqlXpathExtractNumValue($alias, $property) . " " .
-            $operator . " " .
-            $literalOperand->getLiteralValue();
+        if ($this->platform instanceof MySqlPlatform) {
+            switch ($operator) {
+                case '=':
+                case '!=':
```
dantleech: Ignore this switch case..

dantleech:
@dbu @lsmith77 what do you think to this approach?

lsmith77:
if it fixes the issue, then its fine to me. however I wonder if we have the issue with SQLite/PostgreSQL as well? I guess a test case makes sense ..

dantleech:
I imagine we will, lets add a test case and see how it goes.

dbu:
seems to only work for postgres. and we seem to have some issue with reloading fixtures.

dantleech:
Thats odd, I am pretty sure I developed it in MySQL. Will have another look soon regardless.

dbu:
i added this to the 1.2 milestone, as lukas says this is important for sulu. is it?

dantleech:
yeah, it would be very useful /cc @watcherjohannes

dantleech:
ok this should pass now, although I suspect we still have a flaky test lurking in master. (`/container/idExample` not found).

This PR also makes me think that we should handle the query building with events, or at least provide a more structured way of specializing queries for different drivers. Maybe something to think about.

dbu:
the flaky test is noted in #239

dbu:
mysql reliably fails. is this related to what you are doing or not?

dantleech:
yeah sorry -- just found out I was only testing locally on sqlite. debugging MySQL now.

dantleech:
ok, builds now passing apart from the flaky sqlite.

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -627,10 +627,28 @@ public function walkNumComparisonConstraint(QOM\PropertyValueInterface $property
         $alias = $this->getTableAlias($propertyOperand->getSelectorName() . '.' . $propertyOperand->getPropertyName());
         $property = $propertyOperand->getPropertyName();
 
-        return
-            $this->sqlXpathExtractNumValue($alias, $property) . " " .
-            $operator . " " .
-            $literalOperand->getLiteralValue();
+
+        if ($this->platform instanceof MySqlPlatform) {
```
dbu: to make this more readable, could we do?

```
if ($this->platform instanceof MySqlPlatform && in_array($operator, array('=', '!='))) {
    return sprintf(...
}
```

dbu:
great. apart from the readability i think this is good now. i really would merge the if and the switch into one, about the assignment vs direct return i'd prefer the direct return but don't mind too much.

can you then squash the commits please?

dantleech:
Restructured the code. Also found out that `!=` is not valid JCR-SQL2. So have just made it act only on equality for now.

dantleech:
and squashed.

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -627,10 +627,22 @@ public function walkNumComparisonConstraint(QOM\PropertyValueInterface $property
         $alias = $this->getTableAlias($propertyOperand->getSelectorName() . '.' . $propertyOperand->getPropertyName());
         $property = $propertyOperand->getPropertyName();
 
-        return
-            $this->sqlXpathExtractNumValue($alias, $property) . " " .
-            $operator . " " .
-            $literalOperand->getLiteralValue();
+
+        if ($this->platform instanceof MySqlPlatform && $operator == '=') {
```
lsmith77: ``if ($this->platform instanceof MySqlPlatform && $operator == '=') {``

=>

``if ($this->platform instanceof MySqlPlatform && '=' === $operator) {``

dantleech: Does `===` offer any advantage here? Changed anyway.

lsmith77: its slightly faster and of course its type safe. I just find it a good practice to "default" to ``===``

dantleech:
good to merge?

dbu:
great, thanks a lot!

PR #244, opened 2015-02-09
validate namespaces in paths

### LINKED ISSUE #208: validate namespace prefix in nodes before doing lookups (and writes?)

dbu:
fix #208

until a version of phpcr-utils having https://github.com/phpcr/phpcr-utils/pull/145 is used, the extra argument is ignored and no validation really happens.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": ">=5.3.3",
         "doctrine/dbal": "~2.4",
         "phpcr/phpcr": "~2.1.2",
-        "phpcr/phpcr-utils": "~1.2",
+        "phpcr/phpcr-utils": "~1.2,>=1.2.4",
         "jackalope/jackalope": "~1.2.0"
     },
     "provide": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog
 1.2
 ---
 
+* Validate namespaces in node paths and throws NamespaceException on unknown prefixes.
 * Throw Exception when user tries to select either jcr:path or jcr:score
 * The jackalope:init:dbal command now only really executes when the --force
   parameter is given.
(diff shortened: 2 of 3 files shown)


lsmith77:
we also then need to bump the phpcr utils dep requirement

dbu:
not necessarily. php is sloppy enough to just ignore the additional parameter. we just don't validate in that case which seems ok to me. but we can also tag a new dot release of the utils and update the dependency, its probably better (and probably does not matter as when updating phpcr-utils gets updated anyways)

lsmith77:
+1 for tagging a new release and bumping the dep

dbu:
tagged phpcr-utils and updated composer.json . lets see

dbu:
okay, that phpcr-utils validation is wrong, it must not match `/` in prefixes

composer.json
```diff
@@ -19,7 +19,7 @@
         "php": ">=5.3.3",
         "doctrine/dbal": "~2.4",
         "phpcr/phpcr": "~2.1.2",
-        "phpcr/phpcr-utils": "~1.2",
+        "phpcr/phpcr-utils": "fix-validate-namespaces",
```
dbu: needs to become ~1.2,>1.2.4 before merging

dbu:
well. did two mistakes: https://github.com/phpcr/phpcr-utils/pull/146 and i passed the namespace hashmap instead of just the namespace keys to validation. and i fixed a glitch in the api tests: https://github.com/phpcr/phpcr-api-tests/commit/7c90f87bc959b1bdfa9489c3198034fa14163caf

this time i have green locally

dbu:
updated the phpcr-utils dep now that we merged and tagged

dbu:
@lsmith77 good to merge?

PR #245, opened 2015-02-14
Update master from 1.1

dbu:
did not want to push directly to 1.1 but to a branch off 1.1. but i think this is correct. tested it in the cmf sandbox and it works.


lsmith77:
looks green here as well .. so merge?

PR #247, opened 2015-02-18
Test lowest versions

dbu:
replaces #235, now against master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,18 @@ php:
   - hhvm
 
 env:
-  - DB=mysql
-  - DB=pgsql
-  - DB=sqlite
+  - DB=mysql PACKAGE_VERSION=high
+  - DB=mysql PACKAGE_VERSION=low
+  - DB=pgsql PACKAGE_VERSION=high
+  - DB=pgsql PACKAGE_VERSION=low
+  - DB=sqlite PACKAGE_VERSION=high
+  - DB=sqlite PACKAGE_VERSION=low
 
 before_script:
+  - composer selfupdate
   - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
-  - composer self-update
-  - composer update --prefer-source --no-interaction
+  - if [[ "$PACKAGE_VERSION" == "high" ]]; then composer update --prefer-source --no-interaction; fi
+  - if [[ "$PACKAGE_VERSION" == "low" ]]; then composer update --prefer-lowest --prefer-source --no-interaction; fi
   - if [[ "$DB" == "mysql" ]]; then mysql -e "create database phpcr_tests;"; fi
   - if [[ "$DB" == "pgsql" ]]; then psql -c "create database phpcr_tests;" -U postgres; fi
   - php tests/generate_phpunit_config.php $DB
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,21 @@ php:
 
 env:
   - DB=mysql PACKAGE_VERSION=high
-  - DB=mysql PACKAGE_VERSION=low
   - DB=pgsql PACKAGE_VERSION=high
-  - DB=pgsql PACKAGE_VERSION=low
   - DB=sqlite PACKAGE_VERSION=high
-  - DB=sqlite PACKAGE_VERSION=low
+
+
+matrix:
+  allow_failures:
+    - php: hhvm
+    # hhvm will fail until libxml support is added https://github.com/facebook/hhvm/issues/1194
+  include:
+    - php: 5.3.3
+      env: DB=mysql PACKAGE_VERSION=low
+    - php: 5.3.3
+      env: DB=pgsql PACKAGE_VERSION=low
+    - php: 5.3.3
+      env: DB=sqlite PACKAGE_VERSION=low
 
 before_script:
   - composer selfupdate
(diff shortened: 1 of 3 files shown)


dbu:
there is one failure left because something is not an array. it would work with dbal 2.5.0 - shall i bump the version requirement to that? or can we avoid the error on 2.4? i guess it would be nice to not force a too high version requirement on this one to not mess with doctrine orm parallel use.

@lsmith77 do you have an idea if we can do something on our side to avoid that error? it must have worked in previous jackalope versions as we did work with dbal 2.4 at some point...

lsmith77:
no .. I do not see how we can know if the visitors will generate any queries before hand without duplicating a lot of code.

maybe @Ocramius has an idea?

Ocramius:
What's the problem with running latest, if I may ask?

lsmith77:
@Ocramius

the code line that is causing the issue is:
```
foreach ($schema->toSql($connection->getDatabasePlatform()) as $sql) {
```

this results in the following stack trace:
```
1) Jackalope\Tools\Console\Command\InitDoctrineDbalCommandTest::testCommand
array_merge(): Argument #2 is not an array
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php:68
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Table.php:738
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Schema.php:393
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Schema.php:335
/home/travis/build/jackalope/jackalope-doctrine-dbal/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php:96
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:214
/home/travis/build/jackalope/jackalope-doctrine-dbal/vendor/symfony/console/Symfony/Component/Console/Tester/CommandTester.php:66
/home/travis/build/jackalope/jackalope-doctrine-dbal/tests/Jackalope/Tools/Console/InitDoctrineDbalCommandTest.php:77
/home/travis/build/jackalope/jackalope-doctrine-dbal/tests/Jackalope/Tools/Console/InitDoctrineDbalCommandTest.php:85
```

dbu:
@Ocramius we are trying to set up building with the lowest deps to ensure our minimal requirements are not a lie. in this case, it looks like the failing test is indeed telling that you can't run the init command with dbal 2.4

the problem is in CreateSchemaSqlCollector::acceptTable where $this->platform->getCreateTableSQL($table) is not returning the expected array. 2.5 casts that to array to avoid the problem. can we do anything to work around that for 2.4?

Ocramius:
@dbu you need to ask @deeky666, maybe he can backport the bugfix to 2.4.x

Ocramius:
It may be a too big BC break for a patch release tho

dbu:
can we work around it somehow? like make sure that we give the platform something that it returns an array?

on the other hand: would there be a problem if we bump the *minimal* version requirement for dbal to 2.5? could that cause issues for users? otherwise i would say we just do that and push for upgrades.

Ocramius:
> would there be a problem if we bump the minimal version requirement for dbal to 2.5? 

Not really: ORM 2.4 runs also with DBAL 2.5, and better than before :P

dbu:
okay. lets go with 2.5 then

deeky666:
The DBAL patch cannot be backported easily to 2.4 as it was fixed as  part of a new feature. We could/should however fix that issue directly in 2.4 branch.

Ocramius:
@deeky666 we can't change the method return type :-\

deeky666:
@Ocramius it's about [this line](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php#L81) (explicit array cast) as far as I understood. That cast is missing in 2.4.

dbu:
yep, that seems to be the issue. i am bumping to 2.5 now. but maybe 2.4 could use a backport of that fix.

dbu:
@lsmith77 @WouterJ @dantleech good to merge? if so, we are down to one problem before releasing

WouterJ:
Can we please look at backporting the fix to 2.4 again?

As Doctrine ORM 2.5 requires PHP 5.4, having it as dependency in your PHP 5.3 project means that all doctrine library dependencies are downgraded to Doctrine ORM 2.4. This results in an installation of jackalop-doctrine-dbal 1.2 instead of 1.3...

Or should we drop PHP 5.3 support from the CMF bundles?

dbu:
@deeky666 what do you think about backporting the fix you mentioned to dbal 2.4 as that is the last version that supports php 5.3? imho, we should support 5.3 for another couple of months to keep working with the horribly outdated redhat enterprise servers.

dbu:
i can do the pull request if you agree it makes sense. not sure if dbal 2.4 is still maintained at all...

Ocramius:
2.5 still runs on 5.3: no need to backport.

WouterJ:
@Ocramius as soon as someone is going to install Doctrine ORM in its PHP 5.3 project, every doctrine package is going to be downgraded to 2.4 as `doctrine/orm:2.5.x` requires PHP 5.4. The change that someone uses ORM and PHPCR in a project is pretty big, as not every problem can be solved with PHPCR.

Ocramius:
I'm talking about DBAL 2.5

deeky666:
@dbu we cannot really backport that one line as it was added as part of a feature. But sure, go ahead and do a PR against 2.4 branch that fixes the issue. I see no problems with that...

PR #250, opened 2015-03-18
Always initialize the connection if it is closed

### LINKED ISSUE #249: Sqlite: SQLSTATE[HY000]: General error: 1 no such function: CONCAT
Am seeing this when running the CMF RoutingAutoBundle tests. Relaunching a successful build from  1 month ago also gives the same error.

https://travis-ci.org/symfony-cmf/RoutingAutoBundle/jobs/54443252#L450

This is quite strange, the call to register the CONCAT function with sqlite is being executed.

dantleech:
Always initialize if the connection is closed. Fixes #249 

The DoctrineBundle [now closes connections](https://github.com/doctrine/DoctrineBundle/commit/2f8d7263c4c44445b51fc2d9bf71d21f7016347d) on kernel#shutdown, which can cause issues for functional tests.

When the connection is reestablished the registered functions (e.g. CONCAT for sqlite) are lost. This PR will reinitialize if it sees that the connection is closed.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -2600,7 +2600,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
      */
     private function initConnection()
     {
-        if (true === $this->connectionInitialized) {
+        if (true === $this->conn->isConnected() && true === $this->connectionInitialized) {
             return;
         }
 


dbu:
looks good. is there any sort of event system that we could listen to, to change the initialized information in our client when the connection is closed?

dantleech:
Actually yeah, these is a `postConnect` event that might do the trick.

Issue #251, opened 2015-03-24
When copying nodes nodes which the copied node referrers to are not back-referenced.


dantleech:
For example, if you copy a tree which has a node `B` with a reference to node `A` somewhere, then node `B` is not listed in the list of references for node `A`.

dbu:
fixed in #252

PR #252, opened 2015-03-24
Sync references after copy

dantleech:
Sync references afer copy, fixes: https://github.com/jackalope/jackalope-doctrine-dbal/issues/251

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -584,7 +584,12 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
             $resultSetUuids[$row['identifier']] = $row['path'];
         }
 
+        // array of references to remap within the copied tree
         $referenceElsToRemap = array();
+
+        // array references that will need updating in the database
+        $referencesToUpdate = array();
+
         foreach ($rows as $row) {
             $newPath = str_replace($srcAbsPath, $dstAbsPath, $row['path']);
 
@@ -598,13 +603,26 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
             $propsData = array(
                 'stringDom' => $stringDom,
-                'numericalDom' => $numericalDom
+                'numericalDom' => $numericalDom,
+                'references' => array(),
             );
 
             $xpath = new \DOMXpath($stringDom);
-            $referenceEls = $xpath->query('.//sv:property[@sv:type="reference"]');
+            $referenceEls = $xpath->query('.//sv:property[@sv:type="reference" or @sv:type="Reference" or @sv:type="weakreference" or @sv:type="WeakReference"]');
 
+            $references = array();
             foreach ($referenceEls as $referenceEl) {
+                $propName = $referenceEl->getAttribute('sv:name');
+                $values = array();
+                foreach ($xpath->query('./sv:value', $referenceEl) as $valueEl) {
+                    $values[] = $valueEl->nodeValue;
+                }
+
+                $references[$propName] = array(
+                    'type' => PropertyType::valueFromName($referenceEl->getAttribute('sv:type')),
+                    'values' => $values,
+                );
+
                 if (isset($resultSetUuids[$referenceEl->nodeValue])) {
                     $referenceElsToRemap[] = array($referenceEl, $newPath, $row['type'], $propsData);
                 }
@@ -614,6 +632,13 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
             // when copying a node, the copy is always a new node. set $isNewNode to true
             $newNodeId = $this->syncNode(null, $newPath, $row['type'], true, array(), $propsData);
+
+            if ($references) {
+                $referencesToUpdate[$newNodeId] = array(
+                    'path' => $row['path'],
+                    'properties' => $references,
+                );
+            }
 
             $newUuid = $this->nodeIdentifiers[$newPath];
             $uuidMap[$originalUuid] = $newUuid;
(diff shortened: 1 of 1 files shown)


dantleech:
https://github.com/phpcr/phpcr-api-tests/pull/155

dbu:
looks correct to me. @lsmith77 wdyt?

dantleech:
I should probaby add some comments and tidy up the variable names.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -781,16 +806,19 @@ private function syncBinaryData($nodeId, $binaryData)
         }
     }
 
-    private function syncReferences()
+    private function syncReferences($referencesToUpdate = null)
     {
-        if ($this->referencesToUpdate) {
+        $referencesToUpdate = $referencesToUpdate ? : $this->referencesToUpdate;
```
lsmith77: how about making $referencesToUpdate a required parameter while we are at it?

lsmith77:
looks good to me

src/Jackalope/Transport/DoctrineDBAL/Client.php
dantleech: Its set "globally" by `syncNode`

lsmith77: yeah .. but we can just pass it in explicitly in all our calls to syncReferences() .. this would at least get rid of some of the state in this method.

dantleech: done.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -821,6 +853,9 @@ private function syncReferences()
                                     'target_id' => $targetId,
                                 ),
                             );
+
```
lsmith77: extra new lines

dantleech:
ping @dbu how does this look?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -781,16 +810,19 @@ private function syncBinaryData($nodeId, $binaryData)
         }
     }
 
-    private function syncReferences()
+    private function syncReferences($referencesToUpdate)
     {
-        if ($this->referencesToUpdate) {
+        $referencesToUpdate = $referencesToUpdate ? : $this->referencesToUpdate;
```
dbu: i think this is not necessary, as you always pass a value.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -634,6 +659,10 @@ public function copyNode($srcAbsPath, $dstAbsPath, $srcWorkspace = null)
 
             $this->syncNode($this->nodeIdentifiers[$newPath], $newPath, $type, false, array(), $propsData);
         }
+
+        if ($referencesToUpdate) {
```
dbu: the if is kind of redundant, as syncReferences checks this again.

dbu:
@dantleech do we include this in 1.2.0-RC1? can you adress the cleanup comments today? otherwise i just tag RC1 and we can merge this for RC2.

is there an important performance impact by this, by loading the data rather than adjust it with sql statements? though the real performance thing would be to do the actual copy with sql rather than in php, which is clearly out of scope for this PR.

dantleech:
would be great to have it in RC1, will try and have a look today, otherwise RC2

dantleech:
Updated, not sure about the performance impact tbh. will have a quick look

dbu:
actually what would be bad if there is a big impact if there are no references in the copied node. if there are, better an impact than a bug.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -598,13 +603,26 @@ public function copyNode($srcAbsPath, $dstAbsPath, $srcWorkspace = null)
 
             $propsData = array(
                 'stringDom' => $stringDom,
-                'numericalDom' => $numericalDom
+                'numericalDom' => $numericalDom,
+                'references' => array(),
             );
 
             $xpath = new \DOMXpath($stringDom);
             $referenceEls = $xpath->query('.//sv:property[@sv:type="reference"]');
```
dbu: btw, what about weakReference?

dbu:
just re-read it, i think apart from running an xpath against the copied node, there is no impact when there are no references in the node. sounds acceptable.

but while reading, i started wondering whether we are missing weak references in that logic...

dantleech:
Hm. It seems that `sv:type=reference` would never have worked as it should be `sv:type=Reference`. Not sure how that got in there, and the phpcr-api-tests don't seem to work with either `master` or this branch :/ Will continue to look.

Did a benchmark, `5.6s` before and `12.6s` after on a dataset with ~400 nodes with references, havn't tested on a copy without reference yet.

lsmith77:
in terms of performance the case without references is more interesting. obviously actually copying will reduce performance than incorrectly doing nothing.

dantleech:
Updated.

The PHPCR tests are using lowercased property type names, whereas the actual implementation is using camel case, so here I check for both -- any better ways? Otherwise this is good with the `phpcr-api-tests` now.

Benchamrked copying a content tree of `~200` nodes, which has a few references, but not many (as opposed to the route tree that I copied earlier which has references in each node). There was no significant impact:

````
This branch:
3.476
3.484
3.515

Master:
3.46
3.428
3.440
````

dbu:
squashed and merged. thanks a lot dan!

i assume that the difference between camel case and lower case comes from jackrabbit using lowercase internally. we should probably check what the node import / export is doing and standardize on that.

PR #254, opened 2015-04-01
Fix for The ArrayObject Bug in the CachedClient

### LINKED ISSUE #253: ArrayObject-Bug in Jackalope\Transport\DoctrineDBAL\Client.php
Since namespace is an ArrayObject now, there is the need to cast it to an array...which is done in:

   /**
     * {@inheritDoc}
     */
    public function getNamespaces()
    {
        return (array) $this->getNamespacesObject();
    }

But this is overriden by the CachedClient "getNamespaces"-Method.

Which just checks if namespaces is filled:

 /**
     * {@inheritDoc}
     */
    public function getNamespaces()
    {
        if (empty($this->namespaces)) {
            $cacheKey = 'namespaces';
            $result = $this->caches['meta']->fetch($cacheKey);
            if ($result) {
                $this->namespaces = $result;
            } else {
                $this->namespaces = parent::getNamespaces();
                $this->caches['meta']->save($cacheKey, $this->namespaces);
            }
        }
        return $this->namespaces;
    }

And if so, it justs returns its...but it is not casted there....so the engine throws a warning.

Easy fix is to cast it in the CachedClient as well...but a less "doubled" solution would be better i think. :)

Like casting it in "getNamespacePrefixes()" in the Client-Class.

ARoddis:
I had to change the visibility of the "getNamespacesObject" Method. I hope that is ok.

fix #253

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -522,7 +522,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
      *
      * @return \ArrayObject
      */
-    private function getNamespacesObject()
+    protected function getNamespacesObject()
     {
         if ($this->namespaces->count() === 0) {
             $query = 'SELECT * FROM phpcr_namespaces';
--- a/tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
+++ b/tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
@@ -1,0 +1,64 @@
+<?php
+namespace Jackalope\Transport\DoctrineDBAL;
+
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Jackalope\Test\TestCase;
+use PHPCR\PropertyType;
+use PHPCR\Util\NodeHelper;
+use PHPCR\Util\PathHelper;
+
+
+class CachedClientTest extends TestCase
+{
+    /**
+     * @var Client
+     */
+    private $transport;
+
+    /**
+     * @var \Jackalope\Repository
+     */
+    private $repository;
+
+    /**
+     * @var \Jackalope\Session
+     */
+    private $session;
+
+    public function setUp()
+    {
+        static $initialized = false;
+        parent::setUp();
+
+        $conn = $this->getConnection();
+        $options = array('disable_fks' => $conn->getDatabasePlatform() instanceof SqlitePlatform);
+        $schema = new RepositorySchema($options, $conn);
+        $tables = $schema->getTables();
+
+        foreach ($tables as $table) {
+            $conn->exec('DELETE FROM ' . $table->getName());
+        }
+
+        $this->transport = new \Jackalope\Transport\DoctrineDBAL\CachedClient(new \Jackalope\Factory(), $conn);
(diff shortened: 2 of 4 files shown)


dbu:
looks good to me. visiblity change is ok, we default to private but when there is a good reason protected is fine.

@dantleech can you check this as well please?

lsmith77:
ping

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
+namespace Jackalope\Transport\DoctrineDBAL;
+
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Jackalope\Test\TestCase;
+use PHPCR\PropertyType;
+use PHPCR\Util\NodeHelper;
+use PHPCR\Util\PathHelper;
+
+
```
dantleech: remove extra line, missing header.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -522,7 +522,7 @@ public function getNamespaces()
      *
      * @return \ArrayObject
      */
-    private function getNamespacesObject()
+    protected function getNamespacesObject()
```
dantleech: Does this need to be `protected`?

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -0,0 +1,64 @@
...
+    private $repository;
+
+    /**
+     * @var \Jackalope\Session
+     */
+    private $session;
+
+    public function setUp()
+    {
+        static $initialized = false;
```
dantleech: Is this used by anything?

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $tables = $schema->getTables();
+
+        foreach ($tables as $table) {
+            $conn->exec('DELETE FROM ' . $table->getName());
+        }
+
+        $this->transport = new \Jackalope\Transport\DoctrineDBAL\CachedClient(new \Jackalope\Factory(), $conn);
+        $this->transport->createWorkspace('default');
+
+        $this->repository = new \Jackalope\Repository(null, $this->transport);
```
dantleech: @dbu is there nothing existing that does this bootstrapping?

dantleech:
Other than my comment this is good for me.

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
dbu: looks like a copy-paste that is also present but not used elsewhere.

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
dbu: aparently not. i will do a bit of refactoring of the tests i think.

dbu:
thank you very much @ARoddis - i did the codestyle fixes and opened a new PR which i will merge as soon as its green. i will quickly cleanup the test initialization also.

PR #255, opened 2015-04-04
[WIP] Versioning

danrot:
This PR adds versioning as specified in http://www.day.com/specs/jcr/2.0/15_Versioning.html to this transport layer. But most of the work is done in https://github.com/jackalope/jackalope/pull/280.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "doctrine/dbal": "~2.5,>=2.5.0",
         "phpcr/phpcr": "~2.1.2",
         "phpcr/phpcr-utils": "~1.2,>=1.2.4",
-        "jackalope/jackalope": "~1.2.0"
+        "jackalope/jackalope": "dev-versioning as 1.3.0"
     },
     "provide": {
         "jackalope/jackalope-transport": "1.1.0"
@@ -39,5 +39,11 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/jackalope.git"
+        }
+    ]
 }
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -48,9 +48,6 @@
 
             <exclude>
                 <!-- ignore whats not implemented yet -->
-                <directory>vendor/phpcr/phpcr/src/PHPCR/Version</directory>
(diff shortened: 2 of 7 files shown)


dbu:
you should change composer.json to use `"jackalope/jackalope": "dev-versioning as 1.3.0"` and add a section to make it find your fork:
```
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/danrot/jackalope.git"
        },
```

dbu:
@danrot with all the things you do, i wonder if it would make sense to start doing pull requests against your versioning branch, activating the tests as you implement the features. then we can for example be sure there are no regressions. wdyt? i could watch your fork to get notifications. that way, the final review should be a lot easier...

danrot:
@dbu I generally like the idea, but the problem is that it is hard for me to create isolated pull requests that make sense, since I don't know the code base too well. Basically I am just editing files as the need arises to :-)

Or how would you have imagined that? Should we agree on some steps, and if e.g. the checkin is fully working this is a separate PR?

dbu:
this PR against master can only be merged when basically everything is done. my idea would be to have one branch - maybe even here in the official repository - against which you do the pull requests. once a feature is good, we can merge it. this can be as little as making just a single additional test pass, or a refactoring of something you need, or a whole part like checkin working, or frozen node features or the like.

as said, the idea would be that more and more tests get enabled and the versioning branch is testing green (by skipping the phpcr-api-tests for things that are not yet done)

dbu:
first step would be to adjust https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/tests/ImplementationLoader.php so it skips all tests or cases that are not succeeding yet.

i think it will also be more gratifying when we can merge something from time to time, its clearer to see progress :-)

danrot:
Ok, then I will first push everything I have now, and then we should try to agree that what I have been until now is good, otherwise you will have to review everything at the end again. Do you agree?

dbu:
yes. i propose you close this PR, then create a new PR with your branch against the `versioning` branch i just created in this repository. adjust the implementation loader to only run the tests that are green, and i will review what we have so far and then we merge that. from then on you can do pull requests against `versioning` in this repository

danrot:
Ok, since I already started the checkout method on the hackday I would like to make the tests passing affected by this first. Then I will create new PRs. Can you also create a versioning branch in jackalope?

dbu:
agreed for wrapping up checkin.

i created the branch in jackalope/jackalope now.

PR #257, opened 2015-04-05
Cleanup functional tests

dbu:
If you want to improve the test coverage, please start from this branch @ARoddis

One idea i had was that CachedClientTest and ClientTest could share many of the tests, so that we usually test the behaviour with caching automatically as well and see if anything breaks.

### DIFF
--- a/tests/Jackalope/Test/FunctionalTestCase.php
+++ b/tests/Jackalope/Test/FunctionalTestCase.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Jackalope\Test;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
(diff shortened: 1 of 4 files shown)


ARoddis:
That was my first idea. I extended the normal TestClient. So it would check for changed behaivior in the but as it would alter teststats strongly, i wasnt bold enought. ^^

But like you suggested it was my first idea aas well. 

If its ok ill "rredo" it and start from there. :)

dbu:
i am not sure if all tests in ClientTest make sense to redo, or if they even work. maybe have a look and if they do, extending ClientTest would sound good to me. otherwise split ClientTest into an abstract class BaseClientTest that holds the common test cases and a ClientTest | CachedClientTest with extra tests.

i will just merge this here now, then you can start from master. feel free to refactor, we have pull requests to review things ;-)

PR #258, opened 2015-04-09
dbal fix is backported

dbu:
follow up of #247 

this should fail for now, so that we can compare. then we can try the branch https://github.com/doctrine/dbal/compare/2.4...dbu:backport-creation-fix-to-2.4?expand=1 and it should no longer fail, then do a PR with that branch...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "doctrine/dbal": "~2.5,>=2.5.0",
+        "doctrine/dbal": ">=2.4.5,<3.0.x-dev",
         "phpcr/phpcr": "~2.1.2",
         "phpcr/phpcr-utils": "~1.2,>=1.2.4",
         "jackalope/jackalope": "~1.2.0"


dbu:
updated with my branch, lets see if this fixes the problem.

composer.json
```diff
@@ -15,9 +15,15 @@
...
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/dbu/dbal.git"
+        }
+    ],
     "require": {
         "php": ">=5.3.3",
-        "doctrine/dbal": "~2.5,>=2.5.0",
+        "doctrine/dbal": "dev-backport-creation-fix-to-2.4 as 2.4.99",
```
lsmith77: I think othe dot in the branch name causes issues for packagist.org

/cc @Seldaek

dbu: i will remove this once dbal has merged things.

but as far as i can see, all is fine with the name. i stopped all builds but one to see quicker whether things work out.

dbu:
verified that the change really fixes the issue, created https://github.com/doctrine/dbal/pull/835

dbu:
this should be ready to merge. as soon as https://github.com/doctrine/dbal/pull/835 turns up in a 2.4.5 release we can use jackalope-doctrine-dbal with the dbal 2.4 again.

PR #259, opened 2015-04-11
cache ItemNotFoundException in the CachedClient

lsmith77:
dropped the query count from 68 to 2

https://blackfire.io/profiles/compare/66c8148d-478c-4bd5-8864-986bc3a6bc66/graph

now the main overhead remains inside Jackalope core

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -136,10 +136,23 @@ class CachedClient extends Client
 
         $cacheKey = "nodes: $path, ".$this->workspaceName;
         if (isset($this->caches['nodes']) && (false !== ($result = $this->caches['nodes']->fetch($cacheKey)))) {
+            if ('ItemNotFoundException' === $result) {
+                throw new ItemNotFoundException(sprintf('Item "%s" not found in workspace "%s"', $path, $this->workspaceName));
+            }
+
             return $result;
         }
 
-        $node = parent::getNode($path);
+        try {
+            $node = parent::getNode($path);
+        } catch (ItemNotFoundException $e) {
+            if (isset($this->caches['nodes'])) {
+                $this->caches['nodes']->save($cacheKey, 'ItemNotFoundException');
+            }
+
+            throw $e;
+        }
+
         if (isset($this->caches['nodes'])) {
             $this->caches['nodes']->save($cacheKey, $node);
         }


src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -134,10 +134,23 @@ public function getNode($path)
 
         $cacheKey = "nodes: $path, ".$this->workspaceName;
         if (isset($this->caches['nodes']) && (false !== ($result = $this->caches['nodes']->fetch($cacheKey)))) {
+            if ('' === $result) {
+                throw new ItemNotFoundException("Item $path not found in workspace ".$this->workspaceName);
```
dantleech: Would be nice to use `sprintf` here and quote the values, `sprintf('Item "%s" not found in workspace "%s"' ..`

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -134,10 +134,23 @@ public function getNode($path)
...
+
             return $result;
         }
 
-        $node = parent::getNode($path);
+        try {
+            $node = parent::getNode($path);
+        } catch (\Exception $e) {
+            if (isset($this->caches['nodes'])) {
+                $this->caches['nodes']->save($cacheKey, '');
```
dantleech: Would it make more sense to make this a meaningful value? e.g. `__no_item__` and use a constant?

lsmith77: I changed it to ``ItemNotFoundException``

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
lsmith77: also fixed .. though I just copied this from the Client class where we currently do not use sprintf in most places. we should maybe go through it all in a separate PR

dbu:
+1 to merge, makes sense to me.

dbu:
should we also cache them in the jackrabbit client? hm, or should we have done this in ObjectManager instead of the client?

lsmith77:
imho no .. or rather we should look into moving the entire caching into the ObjectManager .. not cache misses specifically. caching cache misses here is no problem here because we also have logic to clear the cache on write etc.

dbu:
hm, good point. but do we have the same issue with jackrabbit client and should do something similar there?

lsmith77:
well we do not have a client side cache for jackrabbit which is why yes, we should look at moving the caching into core of jackalope.

PR #261, opened 2015-05-01
do not fetch columns we do not actually use to build the result

lsmith77:


dbu:
i think this has quite some side effects... lots of failures with the tests

lsmith77:
should be fixed now

PR #262, opened 2015-05-01
use only one query to get from uuid to system id when saving references

dbu:
this is mainly naming cleanup, plus a performance improvement


lsmith77:
looks fine .. one could of course think about adding a method that can fetch multiple node ids at once

dbu:
indeed, but that is a more complex refactoring. we would split that loop in two, with the first loop collecting the ids, then one query, then a second loop to map the ids into the $update array, with a check if something is missing.

PR #272, opened 2015-05-05
Use VersionHandler and adapt tests

danrot:
This PR adapts the tests, so that also already supported versioning tests are being run. It also makes use of the VersionHandler introduced in https://github.com/jackalope/jackalope/pull/285

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "doctrine/dbal": "~2.5,>=2.5.0",
         "phpcr/phpcr": "~2.1.2",
         "phpcr/phpcr-utils": "~1.2,>=1.2.4",
-        "jackalope/jackalope": "~1.2.0"
+        "jackalope/jackalope": "dev-versioning as 1.3.0"
     },
     "provide": {
         "jackalope/jackalope-transport": "1.1.0"
@@ -39,5 +39,11 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/jackalope.git"
+        }
+    ]
 }
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": ">=5.3.3",
         "doctrine/dbal": "~2.5,>=2.5.0",
         "phpcr/phpcr": "~2.1.2",
-        "phpcr/phpcr-utils": "dev-versioning as 1.2.4",
+        "phpcr/phpcr-utils": "~1.2,>=1.2.6",
         "jackalope/jackalope": "dev-versioning as 1.3.0"
     },
     "provide": {
@@ -39,15 +39,5 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    },
-    "repositories": [
-        {
-            "type": "vcs",
-            "url": "https://github.com/danrot/jackalope.git"
-        },
-        {
-            "type": "vcs",
-            "url": "https://github.com/danrot/phpcr-utils.git"
-        }
-    ]
+    }
 }
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": ">=5.3.3",
         "doctrine/dbal": "~2.5,>=2.5.0",
         "phpcr/phpcr": "~2.1.2",
-        "phpcr/phpcr-utils": "~1.2,>=1.2.4",
+        "phpcr/phpcr-utils": "dev-versioning as 1.2.4",
         "jackalope/jackalope": "dev-versioning as 1.3.0"
     },
     "provide": {
@@ -44,6 +44,10 @@
         {
             "type": "vcs",
             "url": "https://github.com/danrot/jackalope.git"
+        },
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/phpcr-utils.git"
         }
     ]
 }
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -2,6 +2,7 @@
 
 namespace Jackalope\Transport\DoctrineDBAL;
 
+use Jackalope\Transport\VersioningInterface;
 use PHPCR\LoginException;
 use PHPCR\NodeType\NodeDefinitionInterface;
 use PHPCR\NodeType\NodeTypeExistsException;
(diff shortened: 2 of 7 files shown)


tests/ImplementationLoader.php
```diff
@@ -54,6 +53,11 @@ protected function __construct(Connection $connection, $fixturePath)
                     'Query\\XPath', // Query language 'xpath' not implemented.
                     'Query\\Sql1', // Query language 'sql' is legacy and only makes sense with jackrabbit
                     'Writing\\CloneMethodsTest', // TODO: Support for workspace->clone, node->update, node->getCorrespondingNodePath
+
+                    // TODO fully implement versioning
```
dantleech: Are we not fully implementing versioning?

danrot: We are, but I and @dbu agreed on trying to merge always parts, and skip the tests which are not succeeding yet. This way it's easier for him to review.

danrot:
@dbu The tests are failing because of the following line: https://github.com/phpcr/phpcr-utils/blob/master/src/PHPCR/Util/ValueConverter.php#L136

This is in the PHPCR-Utils repository, should I just remove this check, and create another PR?

danrot:
I don't know another solution, so I just did it :smiley: 
I hope that this doesn't affect the jackrabbit implementation...

composer.json
```diff
@@ -39,5 +39,15 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/jackalope.git"
```
dbu: remove once https://github.com/jackalope/jackalope/pull/285 is merged.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2467,4 +2491,71 @@ private function initConnection()
...
+     * Check-in item at path.
+     *
+     * @see VersionManager::checkin
+     *
+     * @param string $path absolute path to the node
+     *
+     * @return string path to the new version
+     *
+     * @throws UnsupportedRepositoryOperationException
+     * @throws RepositoryException
```
dbu: please remove this comment and replace with `{@inheritDoc}`

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2467,4 +2491,71 @@ private function initConnection()
...
+    /**
+     * Restore the item at versionPath to the location path
+     *
+     * TODO: This is incomplete. Needs batch processing to avoid chicken-and-egg problems
+     *
+     * @see VersionManager::restoreItem
+     */
+    public function restoreItem($removeExisting, $versionPath, $path)
+    {
+        // TODO: Implement restoreItem() method.
```
dbu: please throw a NotImplementedExeption rather than silently failing.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2467,4 +2491,71 @@ private function initConnection()
...
+     *
+     * @param string $versionPath absolute path to the version node
+     * @param string $versionName The name of the version
+     *
+     * @throws ReferentialIntegrityException
+     * @throws VersionException
+     */
+    public function removeVersion($versionPath, $versionName)
+    {
+        // TODO: Implement removeVersion() method.
```
dbu: also throw NotImplementedException

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1817,13 +1826,24 @@ private function getNodeProcessor()
...
 
         return $this->nodeProcessor;
     }
 
     /**
+     * Returns a handler for the versioning mechanism
+     *
+     * @return VersionHandler
+     */
+    private function getVersionHandler()
```
dbu: do we need a getter for a private property? @lsmith77 wdyt?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2467,4 +2491,45 @@ private function initConnection()
...
+        throw new NotImplementedException();
+    }
+
+    /**
+     * Sets the generic version handler delivered by jackalope
+     * @param VersionHandler $versionHandler
+     */
+    public function setVersionHandler(VersionHandler $versionHandler)
+    {
+        $this->versionHandler = $versionHandler;
```
dbu: this is a classic "optional dependency". i think there is no way to avoid it here. but lets at least add a check `if ($this->versionHandler) { throw new \InvalidArgumentException('Version handler is already set'); }`

tests/Jackalope/Test/Fixture/DBUnitFixtureXML.php
```diff
@@ -414,4 +435,77 @@ protected function ensureTableExists($tableName, $columns)
         return $this;
     }
 
+    private function addVersioningProperties(\DOMDocument $dom, \DOMElement $node, $workspaceName, $id)
```
dbu: @lsmith77 if you have a better idea how we can do this, i would be glad. it is what we came up with at the hackday and while it works, its a bit cumbersome.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dantleech: I would say not.

tests/Jackalope/Test/Fixture/DBUnitFixtureXML.php
lsmith77: looked at a bit just now and do not have a better idea ..

src/Jackalope/Transport/DoctrineDBAL/Client.php
lsmith77: yeah .. lets just add it when we need it

danrot: You are right, just retained because it was a singleton in an earlier stage of development...

danrot:
@dbu The failing tests here are because composer loaded an old version of jackalope/jackalope, and I can't restart the tests...

dbu:
okay, i merged the jackalope/jackalope PR. can you update the dependency please? then the build will automatically be restarted.

danrot:
I fear first also https://github.com/phpcr/phpcr-utils/pull/152 has to be merged...

dbu:
argh. i left that lying around too. will have a look what we can do on
jackalope-jackrabbit side.

danrot:
@dbu Have updated the dependencies since you have merged phpcr/phpcr-utils#152, should be fine to merge now.

PR #274, opened 2015-05-08
Added phpcr benchmarks (POC)

dantleech:
This is just a demonstration of the phpcr benchmarking suite I am developing.

The benchmarking reuses the bootstrapping of the `phpcr-api-tests`.

In reality I guess we wouldn't include the benchmarks in the travis build, unless we scaled them down quite alot.

https://gist.github.com/dantleech/e0fd57f6d3a69a13ed43

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,11 @@ before_script:
   - if [[ "$DB" == "mysql" ]]; then mysql -e "create database phpcr_tests;"; fi
   - if [[ "$DB" == "pgsql" ]]; then psql -c "create database phpcr_tests;" -U postgres; fi
   - php tests/generate_phpunit_config.php $DB
+  - cp $DB.phpunit.xml phpunit.xml
 
-script: php vendor/bin/phpunit -c $DB.phpunit.xml
+script: 
+    - php vendor/bin/phpunit
+    - php vendor/bin/phpbench run vendor/phpcr/phpcr-benchmarks/benchmarks/Query
 
 notifications:
   irc: "irc.freenode.org#jackalope"
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     "require-dev": {
         "psr/log": "~1.0",
         "phpcr/phpcr-api-tests": "2.1.4",
+        "phpcr/phpcr-benchmarks": "~1.0@dev",
         "phpunit/phpunit": "4.7.*",
         "phpunit/dbunit": "~1.3"
     },


.travis.yml
```diff
@@ -35,8 +35,11 @@ before_script:
   - if [[ "$DB" == "mysql" ]]; then mysql -e "create database phpcr_tests;"; fi
   - if [[ "$DB" == "pgsql" ]]; then psql -c "create database phpcr_tests;" -U postgres; fi
   - php tests/generate_phpunit_config.php $DB
+  - cp $DB.phpunit.xml phpunit.xml
```
dantleech: The benchmarking bootstrap (`.phpbench`) reads `phpunit.xml` ten `phpunit.xml.dist`.

lsmith77:
You labeled this PR a "POC" .. but is there any reason not to merge it now already?

dantleech:
Well, I am not sure this would be a good thing to add to the CI build in the long run.

1. There are only a few benchmarks here, and they don't take alot of time, but we would add more, the increased build time could become an issue.
2. I don't know how accurate the absolute values of the benchmarks will be on travis, if we want to compare implementations / branches we need a level playing field.
3. `phpbench` hasn't got any tags yet and is currently under development.

Saying that, travis is a qucik win and would provide interesting relative statistics, but I would rather wait before merging.

lsmith77:
we should limit the benchmarks to a single case per database to keep the build time slow down a but under control

.phpbench.dist
```diff
@@ -0,0 +1,32 @@
...
+}
+
+require_once(__DIR__ . '/tests/bootstrap.php');
+
+$config = new PhpBench\Configuration();
+$config->setPath(__DIR__ . '/vendor/phpcr/phpcr-benchmarks/benchmarks');
+$config->addReport(array(
+    'name' => 'console_table',
+    'memory' => true,
+    'revolutions' => true,
```
dantleech: Added configuration.

dantleech:
I've now tagged 0.1 of PhpBench. These benchmarks take about 100s locally, I can reduce if required.

lsmith77:
seems to fail on php 5.3.3

.phpbench.dist
```diff
@@ -0,0 +1,35 @@
...
+require_once(__DIR__ . '/tests/bootstrap.php');
+
+$config = new PhpBench\Configuration();
+$config->setPath(__DIR__ . '/vendor/phpcr/phpcr-benchmarks/benchmarks');
+$config->addReport(array(
+    'title' => 'Query Benchmarks',
+    'name' => 'console_table',
+    'groups' => array('query'),
+    'aggregate' => 'run',
+    'cols' => array('params', 'mean_time', 'mean_memory_diff', 'mean_rps', 'deviation_min'),
```
dantleech: Move report configuration to `phpcr/phpcr-benchmarks`

.travis.yml
```diff
@@ -30,13 +30,17 @@ matrix:
 before_script:
   - composer selfupdate
   - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.3" ]]; then composer remove phpcr/phpcr-benchmarks --dev; fi
```
dantleech: phpbench does not support PHP 5.3 .. how long will Jackalope support it?

dbu: i would propose to drop it in version 2, to also be able to use more
concise code thanks to the new constructs in 5.4.

i think skipping benchmarks on 5.3 is a perfectly valid solution for now.

.travis.yml
dantleech: Add a comment in the travis fiule.

dantleech: Or I could modify the bootstrapping so this is not necessary,

dantleech:
Updated. The query benchmarks now use a real, obfuscated, dump of a website (2300 nodes) and I added a full tree traversal. There are still some issues.

lsmith77:
nice! you already support exporting to XML, right? so now the next big thing would be collecting the data and tracking changes

dantleech:
yeah, but the absolute timing values would be of little value as the containers do not perform consistently. it might be worth it for recording relative metrics and memory usage however.

Ultimately I think there should be a separate repo which can benchmark all of the implementations side-by-side in a consistent environment. (and I will add a feature to phpbench to generate reports from different XML dumps).

lsmith77:
good point .. I guess a trend should be visible none the less. Also relative changes within the tests indeed should also be noticeable, ie. when some tests (f.e. all sorting tests) suddenly get relatively slower compared to everything else.

dantleech:
I guess this could be merged now. I have restriced the benchmarks to 2 iterations per subject, and it is adding 2 minutes to the tests. If that is too much we can limit to 1 iteration (but have no idea about stability at all), or skim off some of the more time-consuming benchmarks.

lsmith77:
should we then move https://github.com/dantleech/phpcr-benchmarks to the phpcr org and do a stable release first?

dantleech:
Yeah sure.. will do.

dantleech:
Added https://github.com/phpcr/phpcr-benchmarks and also https://github.com/phpcr/phpcr-implementation-benchmarks which will run all the implementations side-by-side and generate comparative reports.

lsmith77:
can you also do a stable release of phpcr-benchmark? then again phpbench is still at 0.3 ..

dantleech:
Maybe just tagging to `0.1` would be enough for now and depend on `~0.1`.

dantleech:
Updated to latest version of PHPBench: https://github.com/phpcr/phpcr-benchmarks/blob/master/config/phpbench.json

PR #275, opened 2015-05-09
Fix command with --force and --drop

WouterJ:
#241 was not backported, while the `--force` option was backported. This resulted in errors.

Meanwhile, I also fixed the script to also work with doctrine/dbal 2.4 (in which case the `TableExistsException` did not yet exists).

### DIFF
--- a/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
+++ b/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
@@ -8,6 +8,8 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Exception\TableNotFoundException;
 
 use Jackalope\Transport\DoctrineDBAL\RepositorySchema;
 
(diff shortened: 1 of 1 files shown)


WouterJ:
@dbu can you please merge 1.1 into master (so we have support for dbal 2.4 in master too?)

src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
```diff
@@ -69,16 +71,30 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-                            $output->writeln($e->getMessage());
+                try {
+                    foreach ($schema->toDropSql($connection->getDatabasePlatform()) as $sql) {
+                        if (true === $input->getOption('dump-sql')) {
+                            $output->writeln($sql);
+                        } else {
+                            try {
+                                $connection->exec($sql);
+                            } catch (\Exception $e) {
+                                $output->writeln($e->getMessage());
```
dbu: ups, did we really want to merge this like this? why would we then even care about catching the DBALException below?

dbu:
there is a merge conflict with master. i can port the changes, but guess we first need to sort out if you intended the code as it is now.

src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
WouterJ: Hmm, indeed. Seems like this PR does something wrong and should be revered. However, we should change the try..catch statement in master to be the same as this one.

PR #280, opened 2015-05-27
improve caching

lsmith77:
also cache queries, remove redundant code, refactor for better efficiency when nodes are not cached

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\Connection;
 use Jackalope\NotImplementedException;
 use Jackalope\FactoryInterface;
 use Jackalope\Node;
+use Jackalope\Query\Query;
 
 use PHPCR\ItemNotFoundException;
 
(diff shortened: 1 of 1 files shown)


src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -58,9 +72,6 @@ public function deleteWorkspace($name)
 
         $this->caches['meta']->delete('workspaces');
         $this->caches['meta']->delete("workspace: $name");
-        if (isset($this->caches['nodes'])) {
```
dbu: why do we not need to clear the node cache anymore?

lsmith77: I think removing this was a mistake.

lsmith77: fixed

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -339,27 +373,11 @@ public function moveNodeImmediately($srcAbsPath, $dstAbsPath)
     /**
      * {@inheritDoc}
      */
-    public function reorderNodes($absPath, $reorders)
```
lsmith77: this method no longer exists in the Client

lsmith77:
ok .. fixed a few issues and now I think its ready to be merged

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
dbu: i think it was renamed to reorderChildren. we might need to capture that in order to invalidate caches. actually it would probably be enough to invalidate the cache for just the one node that is passed in, as reordering has no other effects than on the parent node of the reordered nodes.

dbu:
:+1:  looks good. the only thing we miss is about the reordering - i think we need to invalidate the cache of the node being reordered.

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
lsmith77: fixed .. I added a ``clearNodeCache()`` method .. there are a few places where we could also use it if we would have access to a ``Node`` instance which I have noted in comments.

dbu:
cool. so those comments we can handle some other day, right? ready to merge?

should this go into the current version or do we start the next minor version?

lsmith77:
imho next minor version .. I can bump the composer.json for this

dbu:
thanks a lot for this work!

dbu:
i bumped the version directly to master.

PR #277, opened 2015-05-19
Enabling tests and adapting fixtures for version history

danrot:
This PR is based on #272

PR #282, opened 2015-05-30
Enabling tests and adapting fixtures for version history

danrot:
Created because #277 had wrong base branch

This PR is still based on #272

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,5 +39,11 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/jackalope.git"
+        }
+    ]
 }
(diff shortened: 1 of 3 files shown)


dbu:
can you please rebase this on the versioning branch now so we can see what is new? there seems to be a conflict as well, unfortunately.

danrot:
@dbu done

danrot:
@dbu I can see that there are some changes because of the auto formatter in IntelliJ... Should I try to revert them?

composer.json
```diff
@@ -39,5 +39,11 @@
         "branch-alias": {
             "dev-master": "1.2-dev"
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/danrot/jackalope.git"
```
dbu: this can go away when https://github.com/jackalope/jackalope/pull/286 is merged, right?

dbu:
hm, that autoformatter seems to have a different code style then we where using. if you could undo the style changes that would help - its hard to see here what really changed.

composer.json
danrot: exactly

dbu: and that PR is merged, can you revert the changes on this file please?

danrot:
@dbu Now it should be a lot better :-)

Issue #284, opened 2015-06-17
\Jackalope\Transport\DoctrineDBAL\Client::$namespaces is not object


Tetragramat:
Hi, 
I'm using jackalope/jackalope-doctrine-dbal 1.2.1.
I got sometimes this error:
```
Call to a member function count() on a non-object
 IN FILE {myProjectDir}/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
 ON LINE 103
```
That mean \Jackalope\Transport\DoctrineDBAL\Client::$namespaces is not object.
I did some research and found this: 
https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/src/Jackalope/Transport/DoctrineDBAL/Client.php#L541

So if you call \Jackalope\Transport\DoctrineDBAL\CachedClient::getNamespacesObject a second time, then error appears.

Sorry for my english.

lsmith77:
Can you look into providing a fix for this? It should be done on the 1.2 branch so that we can release a bug fix release for this.

lsmith77:
ok .. working on this ..

lsmith77:
can you test https://github.com/jackalope/jackalope-doctrine-dbal/pull/286 ?

Tetragramat:
I will try test it, but it would take some time, because I cannot find when it do that.
This error apears actually only on production server and so I know only when that error was, not what for me anonymous user did.

lsmith77:
fixed via https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.2.2

PR #286, opened 2015-06-18
Cached client fixes

lsmith77:
fix for #284

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -94,25 +94,25 @@ class CachedClient extends Client
     }
 
     /**
-     * Return the namespaces of the current session as a referenceable ArrayObject.
-     *
-     * @return \ArrayObject
+     * {@inheritDoc}
      */
-    protected function getNamespacesObject()
+    public function getNamespaces()
     {
-        if ($this->namespaces->count() === 0) {
-            $cacheKey = 'namespaces';
-            $result = $this->caches['meta']->fetch($cacheKey);
-            if ($result) {
-                $this->namespaces = $result;
-            } else {
-                $this->namespaces = parent::getNamespacesObject();
-
-                $this->caches['meta']->save($cacheKey, $this->namespaces);
-            }
+        if ($this->namespaces instanceof \ArrayObject) {
+            return parent::getNamespaces();
         }
 
-        return $this->namespaces;
+        $cacheKey = 'namespaces';
+        $result = $this->caches['meta']->fetch($cacheKey);
+        if ($result) {
+            $this->setNamespaces($result);
+        } else {
+            $result = parent::getNamespaces();
+
+            $this->caches['meta']->save($cacheKey, $result);
+        }
+
+        return $result;
     }
 
     /**
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -84,13 +84,13 @@ class CachedClient extends Client
     public function getNodeTypes($nodeTypes = array())
     {
         $cacheKey = 'nodetypes: '.serialize($nodeTypes);
-        $nodeTypes = $this->caches['meta']->fetch($cacheKey);
-        if (!$nodeTypes) {
-            $nodeTypes = parent::getNodeTypes($nodeTypes);
-            $this->caches['meta']->save($cacheKey, $nodeTypes);
+        $result = $this->caches['meta']->fetch($cacheKey);
+        if (!$result) {
+            $result = parent::getNodeTypes($nodeTypes);
+            $this->caches['meta']->save($cacheKey, $result);
         }
 
-        return $nodeTypes;
+        return $result;
     }
 
     /**
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -130,14 +130,30 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     private $checkLoginOnServer = true;
 
     /**
-     * @var \ArrayObject
+     * Using an ArrayObject here so that we can pass this into the NodeProcessor by reference more elegantly
+     *
+     * @var null|\ArrayObject
      */
     protected $namespaces;
 
     /**
-     * @var array The namespaces at initial state, in case of rollback.
+     * @var null|array The namespaces at initial state when making changes to the namespaces, in case of rollback.
      */
-    private $originalNamespaces = array();
+    private $originalNamespaces;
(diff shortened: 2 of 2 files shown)


lsmith77:
@dantleech can you review this? I wonder why we did things so complicated initially

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -84,29 +84,27 @@ protected function workspaceExists($workspaceName)
...
-     * Return the namespaces of the current session as a referenceable ArrayObject.
-     *
-     * @return \ArrayObject
+     * {@inheritDoc}
      */
-    protected function getNamespacesObject()
+    public function getNamespaces()
     {
-        if ($this->namespaces->count() === 0) {
+        if (!isset($this->namespaces)) {
```
dantleech: Does this effectively check if `namespaces` is null? If so I think `null !== $this->namespaces` might be clearer.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -514,32 +513,21 @@ private function getNamespacePrefixes()
...
                 NamespaceRegistryInterface::PREFIX_MIX => NamespaceRegistryInterface::NAMESPACE_MIX,
                 NamespaceRegistryInterface::PREFIX_XML => NamespaceRegistryInterface::NAMESPACE_XML,
                 NamespaceRegistryInterface::PREFIX_SV => NamespaceRegistryInterface::NAMESPACE_SV,
-            ));
+            );
 
-            foreach ($data as $row) {
-                $this->namespaces[$row['prefix']] = $row['uri'];
-            }
+            $this->namespaces+= $namespaces;
```
dantleech: `namespaces += `

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
lsmith77: agreed .. though ``null === $this->namespaces`` is then the correct way

src/Jackalope/Transport/DoctrineDBAL/Client.php
lsmith77: will fix

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -514,32 +513,21 @@ private function getNamespacePrefixes()
...
 
             $this->originalNamespaces = $this->namespaces;
```
dantleech: If we change the namespaces, we also need to reset the node processor as the namespaces are not passed by reference IIRC. Not sure if that effects this PR.

lsmith77: good point .. maybe this is why we used \ArrayObject previously? we could of course also introduce a setter into the NodeProcessor so that we can update the namespaces.

lsmith77: actually no .. it seems like the issue you note should have existed before as well .. since we used ``getNamespaces() during the construction an not ``getNamespaceObject()``

dantleech: this was why we used \ArrayObject originally. From what I can see the array object should always be the same instance in the current code.

lsmith77: no .. see https://github.com/jackalope/jackalope-doctrine-dbal/pull/286/files#diff-a3a0165ed79ca1ba3513ec5ecd59ec56L517 .. we cast to an (array) inside ``getNamespaces()``

lsmith77: maybe the better approach would be to not even pass in the namespaces into the nodeprocessor but instead get the namespaces from the client when needed

dantleech: yeah, but I think that makes a dependency mess. the node processor shouldn't know about the client in an ideal world. The namespaces should be held in an object and there should be only one instance of it internally.

`getNamespaces` is the public API, if we use only `getNamespaceObject` internally that should work? (at least that was the original plan).

lsmith77:
@dantleech ok .. rewrote things based on your feedback

lsmith77:
maybe we should only create the originalNamespaces backup when we actually modify the namespaces so that we know if we even need to bother doing anything in commit/rollback if we actually messed with the namespaces.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -524,30 +530,38 @@ public function getNamespaces()
...
     }
 
     /**
+     * Set the namespaces property while also storing a backup into originalNamespaces
+     *
+     * @param array $namespaces
+     */
+    protected function setNamespaces(array $namespaces)
+    {
+        $this->namespaces = new \ArrayObject($namespaces);
```
dantleech: I think this should be `$this->namespaces->exchangeArray($namespaces);` so that we keep the same instance, if I understand this correctly.

lsmith77: this was taken care of already, since I was using this method in case of a rollback and not this setter. but I have rewritten things to be more clear.

dbu:
fine for me

lsmith77:
@dantleech any more feedback from your side?

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -84,35 +84,35 @@ protected function workspaceExists($workspaceName)
     public function getNodeTypes($nodeTypes = array())
     {
         $cacheKey = 'nodetypes: '.serialize($nodeTypes);
-        $nodeTypes = $this->caches['meta']->fetch($cacheKey);
-        if (!$nodeTypes) {
-            $nodeTypes = parent::getNodeTypes($nodeTypes);
-            $this->caches['meta']->save($cacheKey, $nodeTypes);
+        $result = $this->caches['meta']->fetch($cacheKey);
```
lsmith77: this is a bug unrelated to the namespace stuff ..

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2358,8 +2381,22 @@ public function getSupportedQueryLanguages()
...
+            if ($this->namespaces[$prefix] === $uri) {
+                return;
+            }
+
+            if (isset($this->coreNamespaces[$prefix])) {
+                throw new NamespaceException("Cannot overwrite JCR core namespace prefix '$prefix' to a new uri '$uri'.");
+            }
+        }
+
+        if (!$this->namespaces instanceof \ArrayObject) {
```
lsmith77: we have the same 2 if statements in ``unregisterNamespace()`` .. not sure though if it makes sense to move that into a separate method.

PR #288, opened 2015-06-27
Use autoloading for PHPCR-API tests

dantleech:
See: https://github.com/phpcr/phpcr-api-tests/pull/161


composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.4",
+        "phpcr/phpcr-api-tests": "dev-clean@dev",
```
dantleech: Remember to revert this.

lsmith77:
have you also tested this with Jackrabbit?

dantleech:
Have just run the tests now and they're fine.

tests/bootstrap.php
```diff
@@ -13,10 +13,6 @@
 $loader->add('Jackalope\\Test', __DIR__);
 $loader->add('Jackalope', __DIR__ . '/../vendor/jackalope/jackalope/tests');
```
dbu: do we still need these lines? should those not be autoloaded as well?

composer.json
dbu: change to 2.1.5 and jackalope/jackalope to 1.2.1 once https://github.com/phpcr/phpcr-api-tests/pull/163 is merged and we tagged a new version

tests/bootstrap.php
dbu: ah, the first one yet, but this one not because the jackalope tests should not be autoloaded for consumers of the library

dantleech: hmm.. but thats what `autoload-dev` is for in composer?

dbu: only for the library itself afaik. i think jackalope-doctrine-dbal does
not use the autoload-dev from its dependencies like jackalope/jackalope

Issue #287, opened 2015-06-23
Array-to-string conversion error when creating reference after deleting nodes (?)


dantleech:
After doing the following I recieve a fatal error:

- Purge the workspace
- Do NOT save the session
- Create 2 nodes
- Make the second node reference the first
- Save the session

````
PHPCR\RepositoryException: Error inside the transport layer: Array to string conversion

/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope/src/Jackalope/ObjectManager.php:841
/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope/src/Jackalope/Session.php:444
/home/daniel/www/sulu-io/sulu-document-manager/lib/NodeManager.php:139
/home/daniel/www/sulu-io/sulu-document-manager/lib/Subscriber/Phpcr/GeneralSubscriber.php:129
/home/daniel/www/sulu-io/sulu-document-manager/vendor/symfony/event-dispatcher/EventDispatcher.php:164
/home/daniel/www/sulu-io/sulu-document-manager/vendor/symfony/event-dispatcher/EventDispatcher.php:53
/home/daniel/www/sulu-io/sulu-document-manager/vendor/symfony/event-dispatcher/ContainerAwareEventDispatcher.php:167
/home/daniel/www/sulu-io/sulu-document-manager/lib/DocumentManager.php:125
/home/daniel/www/sulu-io/sulu-document-manager/tests/Functional/DocumentManager/QueryBuilderTest.php:36

Caused by
PHPUnit_Framework_Error_Notice: Array to string conversion

/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:846
/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:2525
/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope/src/Jackalope/ObjectManager.php:836
/home/daniel/www/sulu-io/sulu-document-manager/vendor/jackalope/jackalope/src/Jackalope/Session.php:444
/home/daniel/www/sulu-io/sulu-document-manager/lib/NodeManager.php:139
[...]

PR #292, opened 2015-08-03
make it possible to create and reference nodes within a single transaction

lsmith77:
see https://github.com/phpcr/phpcr-api-tests/pull/165

while investigating #287 I stumbled over this issue.

according to http://stackoverflow.com/questions/7474222/jackrabbit-node-getreferences-not-returning-anything it should be possible to create and reference nodes within a single transaction. however it seems like we create the uuid too late and so when we have the Operations we point to the path and not the UUID, which then prevents us from finding out the system id when we create the references.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,14 +20,14 @@
         "doctrine/dbal": ">=2.4.5,<3.0.x-dev",
         "phpcr/phpcr": "~2.1.2",
         "phpcr/phpcr-utils": "^1.2.8",
-        "jackalope/jackalope": "~1.2.2"
+        "jackalope/jackalope": "~1.2.4"
     },
     "provide": {
         "jackalope/jackalope-transport": "1.1.0"
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.9",
+        "phpcr/phpcr-api-tests": "2.1.10",
         "phpunit/phpunit": "4.7.*",
         "phpunit/dbunit": "~1.3"
     },
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -600,16 +600,16 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
         PathHelper::assertValidAbsolutePath($dstAbsPath, true, true, $this->getNamespacePrefixes());
 
-        $srcNodeId = $this->getSystemIdForNodePath($srcAbsPath, $srcWorkspace);
+        $srcNodeId = $this->getSystemIdForNode($srcAbsPath, $srcWorkspace);
         if (!$srcNodeId) {
             throw new PathNotFoundException("Source path '$srcAbsPath' not found");
         }
 
-        if ($this->getSystemIdForNodePath($dstAbsPath)) {
+        if ($this->getSystemIdForNode($dstAbsPath)) {
             throw new ItemExistsException("Cannot copy to destination path '$dstAbsPath' that already exists.");
         }
 
-        if (!$this->getSystemIdForNodePath(PathHelper::getParentPath($dstAbsPath))) {
+        if (!$this->getSystemIdForNode(PathHelper::getParentPath($dstAbsPath))) {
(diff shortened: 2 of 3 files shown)


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1449,52 +1449,34 @@ public function getNodes($paths)
...
+     * Get the database primary key for node.
      *
-     * @param string $path          Path of the node
+     * @param string $identifier    Path of the identifier
      * @param string $workspaceName To overwrite the current workspace
      *
      * @return bool|string The database id
      */
-    private function getSystemIdForNodePath($path, $workspaceName = null)
+    private function getSystemIdForNode($identifier, $workspaceName = null)
```
lsmith77: I have merged this method with ``getSystemIdForNodeUuid()`` so that it automatically determines if its UUID or not ..

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -875,7 +875,7 @@ private function syncReferences($referencesToUpdate)
             foreach ($toUpdate as $nodeId => $references) {
                 foreach ($references['properties'] as $name => $data) {
                     foreach ($data['values'] as $value) {
-                        $targetId = $this->getSystemIdForNodeUuid($value);
+                        $targetId = $this->getSystemIdForNode($value);
```
lsmith77: when we create the a node we are referencing to in the same session ``$value`` is the path and not the UUID causing the below exception to be triggered.

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
```diff
@@ -525,4 +525,15 @@ public function testCopy()
         $this->assertEquals($child1['props'], $child2['props']);
         $this->assertEquals($child1['numerical_props'], $child2['numerical_props']);
     }
+
+    public function testCreateReferences()
```
lsmith77: should maybe be moved to phpcr-api-tests?

lsmith77:
it seems like we didn't use to allow this at all (https://travis-ci.org/jackalope/jackalope-doctrine-dbal/jobs/73886355#L251):
```
1) Jackalope\Transport\DoctrineDBAL\ClientTest::testCreateReferences
PHPCR\ValueFormatException: Node /child2 must be persisted before being referenceable
```

but I think it should be allowed and therefore this fix is correct ..

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1449,52 +1449,34 @@ public function getNodes($paths)
...
-        return (boolean) $this->getSystemIdForNodePath($path, $workspaceName);
+        return (boolean) $this->getSystemIdForNode($path, $workspaceName);
     }
 
     /**
-     * Get the database primary key for node at path.
+     * Get the database primary key for node.
      *
-     * @param string $path          Path of the node
+     * @param string $identifier    Path of the identifier
```
dbu: "Path or uuid of the node" maybe?

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
dbu: yes, agree that this should go into phpcr-api-tests

dbu:
oh, i guess so. but then we should try to fix this on jackalope-jackrabbit as well if at all possible. each thing that only works with one implementation reduces the interchangeability of the phpcr implementations

lsmith77:
I would expect it to just work with Jackrabbit .. but yeah we should confirm this.

composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.5",
+        "phpcr/phpcr-api-tests": "dev-create_and_reference_nodes_in_transaction",
```
lsmith77: change to 2.1.6 once https://github.com/phpcr/phpcr-api-tests/pull/165 is merged and tagged

dbu:
does the change in https://github.com/jackalope/jackalope/pull/297 not affect this PR? maybe just to be on the safe side, can you include the jackalope/jackalope branch here as well?

PR #297, opened 2015-08-26
fix copy children binary nodes data

rederrik:
Fixed a bug with copying binary data of children nodes while doing Transport:copy.
Had also add a check in getBinaryStream() to make sure requested stream really exists.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.6",
+        "phpcr/phpcr-api-tests": "dev-master",
         "phpunit/phpunit": "4.7.*",
         "phpunit/dbunit": "~1.3"
     },
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -2187,7 +2187,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
     /**
      * {@inheritDoc}
-     * @throws RepositoryException then no binary data found
+     * @throws RepositoryException when no binary data found
      */
     public function getBinaryStream($path)
     {
@@ -2205,7 +2205,6 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         if (count($data) === 0) {
             throw new RepositoryException('No binary data found in stream');
         }
-
 
         $streams = array();
         foreach ($data as $row) {
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -695,7 +695,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
                 '   SELECT ?, b.property_name, ?, b.idx, b.data FROM phpcr_binarydata b WHERE b.node_id = ?';
 
             try {
-                $this->getConnection()->executeUpdate($query, array($newNodeId, $this->workspaceName, $srcNodeId));
+                $this->getConnection()->executeUpdate($query, array($newNodeId, $this->workspaceName, $row['id']));
             } catch (DBALException $e) {
                 throw new RepositoryException("Unexpected exception while copying node from $srcAbsPath to $dstAbsPath", $e->getCode(), $e);
             }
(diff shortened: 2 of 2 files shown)


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2187,6 +2187,7 @@ public function updateNode(Node $node, $srcWorkspace)
 
     /**
      * {@inheritDoc}
+     * @throws RepositoryException then no binary data found
```
dbu: s/then/when

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2201,6 +2202,11 @@ public function getBinaryStream($path)
             array($nodeId, $propertyName, $this->workspaceName)
         );
 
+        if (count($data) === 0) {
+            throw new RepositoryException('No binary data found in stream');
+        }
+
+
```
dbu: extra blank line

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2201,6 +2202,11 @@ public function getBinaryStream($path)
             array($nodeId, $propertyName, $this->workspaceName)
         );
 
+        if (count($data) === 0) {
+            throw new RepositoryException('No binary data found in stream');
```
dbu: what exactly does that mean? and what if i created a binary multivalue property and set it to an empty array? it seems that this is allowed...

or is this meant as sanity check for database inconsistency?

rederrik: Mostly sanity check.
If you run the test w/o this commit it will fail with an error
>fstat() expects parameter 1 to be resource, boolean given

because the node itself has been copied, but its binary data wasn't and BinaryStreamWrapper will get this error on init.

dbu: i think the bug is further down, on line 2228 where we assume that there is exactly one stream when there are not > 1.

or is this only a problem when people copied nodes before your bugfix and now have an incomplete/inconsistent database?
i think the sanity check has to happen in Property::getBinary. that method has the `length` field of the property which in case of multivalue is an array and can thus see when it did not get the number of binary from storage that was to be expected.

does that make sense?

dbu:
the tests have been merged now. we should sort out the error handling. maybe we need a specific test case with a multivalue binary property with an empty array to check for that edge case?

src/Jackalope/Transport/DoctrineDBAL/Client.php
rederrik: I'm not sure, but if it's multivalue shouldn't we get $data = [[],[]] instead of [] when theres simply no data at all?

lsmith77:
can you temporarily change the composer.json file to use dev-master for phpcr-api-tests ?

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: i am not exactly sure either. may i ask you to add another test with a fixture that say multivalue="true" and then has not a single <value> element? that test could go into the Read chapter.

composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.6",
+        "phpcr/phpcr-api-tests": "dev-master",
```
dbu: NOTE: once we tagged phpcr-api-tests we need to change this to 2.1.7

src/Jackalope/Transport/DoctrineDBAL/Client.php
rederrik: Can you show me how to correctly define multivalue property in xml fixture? I've simply added another <sv:value/> and it seems to work fine..

dbu: it should be possible like this: https://github.com/phpcr/phpcr-api-tests/blob/master/fixtures/general/base.xml#L73-L77 . in the case of empty binary, it would look like `<sv:property sv:name="multivalue" sv:type="Binary" sv:multiple="true"/>` (and then hope that the fixtures loading does not choke on that :-) )

rederrik:
Created https://github.com/phpcr/phpcr-api-tests/pull/168 to test some multivalue operations.

lsmith77:
see https://github.com/jackalope/jackalope-doctrine-dbal/pull/299

lsmith77:
superseded by https://github.com/jackalope/jackalope-doctrine-dbal/pull/299

PR #299, opened 2015-10-03
fix binary copy

### LINKED ISSUE #294: Copy binary data for children nodes
Hi,
In \Jackalope\Transport\DoctrineDBAL\Client:copyNode() line 698 we execute [query](https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/src/Jackalope/Transport/DoctrineDBAL/Client.php#L698) to copy binary data, but according to update params we are trying to copy the same data every iteration 
>WHERE b.node_id =  ? [$srcNodeId]

since $srcNodeId is id of the root source node, not current iteration children node.

Is this by mistake or is this intended?

lsmith77:
supersedes #297 
fix #294

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.6",
+        "phpcr/phpcr-api-tests": "2.1.7",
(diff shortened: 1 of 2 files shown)


dbu:
thanks a lot @rederrik !

Issue #302, opened 2015-10-14
Typo in CachedClient:clearNodeCache()?


rederrik:
Hi,
I encountered a problem while trying to reorder phpcr nodes while using doctrine cache. After some research I found what I think is a typo in https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php#L56
There we are trying to access node cache wia $this->caches['node'] while it is stored as $this->caches['nodes'].
I'v submitted PR with a quick fix. Don't know how to test it properly though.

dbu:
ideally we would have a test that triggers this problem so we can show
its fixed. there are some tests in
https://github.com/jackalope/jackalope-doctrine-dbal/tree/master/tests/Jackalope
, if you want to have a look

dbu:
fixed in #303

thanks a lot @rederrik 

if you can build a test around this, its appreciated. but i already merged the actual fix as it makes sense.

rederrik:
Thanks.
I'm not quite sure what do I test here: the cache clearing method? Or an array structure? It's just a typo after all. Maybe we just need some tests for the whole class.
Unfortunately I don't have time to look into it right now, but maybe later.

dbu:
testing the array would not make much sense. but maybe unit tests with a
mock client for those methods that do extra things, and check if the
cache is invalidated when it should be invalidated (meaning the mock
client methods are called or not called, depending on whether the result
is in the cache). but will be quite tricky to do.

the other thing could be to just run the phpcr-api-tests with a cached
client to see if we spot bugs. maybe as a separate build - if we do it
in every build we double the build time of all 12 variants we have atm

PR #303, opened 2015-10-14
fixed a typo in caches array index

rederrik:
See https://github.com/jackalope/jackalope-doctrine-dbal/issues/302


dbu:
@uwej711 might this be related to https://github.com/doctrine/phpcr-odm/pull/661 maybe? just a wild guess, as that one was about reorderings as well...

dbu:
thanks a lot. we should use constants for that it seems...

Issue #304, opened 2015-10-19
Error when using Sqlite


andylibrian:
I am writing a Laravel package that wraps jackalope doctrine dbal.
In my test, I use pdo sqlite driver, and the test failed with following exception:

```
Argument 1 passed to Jackalope\Transport\DoctrineDBAL\Client::registerSqliteFunctions() must be an instance of Doctrine\DBAL\Driver\PDOConnection, instance of PDO given, called in /home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php on line 2631 and defined

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:214

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:2631

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:266

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:429

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope-doctrine-dbal/src/Jackalope/Transport/DoctrineDBAL/Client.php:389

/home/travis/build/ramping/laravel-jackalope-doctrine-dbal/vendor/jackalope/jackalope/src/Jackalope/Repository.php:145
```
Here is the build : https://travis-ci.org/ramping/laravel-jackalope-doctrine-dbal/jobs/86153727

Apparently, it's because of this:

```php
    /**
     * @TODO: move to "SqlitePlatform" and rename to "registerExtraFunctions"?
     *
     * @param PDOConnection $sqliteConnection
     *
     * @return Client
     */
    private function registerSqliteFunctions(PDOConnection $sqliteConnection)
    {
```
After I read these lines:

```php
        // @TODO: move to "SqlitePlatform" and rename to "registerExtraFunctions"?
[...]

PR #306, opened 2015-10-20
Loosen type hint from DBAL PDOConnection to simply PDO object

andylibrian:
Refer to #304

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -28,7 +28,6 @@ use PHPCR\Util\ValueConverter;
 use PHPCR\Util\UUIDHelper;
 use PHPCR\Util\PathHelper;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\PDOConnection;
 use Doctrine\DBAL\DBALException;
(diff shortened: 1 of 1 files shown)


lsmith77:
looks safe to me ..

@dbu wdyt?

dbu:
if this is needed, https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1357 is lying as \PDO is not implementing Connection. @andylibrian if you change the typehint to the \Doctrine\DBAL\Driver\Connection class instead of the PDOConnection, does it work too? that would seem the correct fix.

dbu:
though the sqliteCreateFunction method is not in Connection or PDOConnection itself, but directly in \PDO. so lets go for \PDO. can you adjust the `@param` annotation please, and remove the now unused `use` statement?

andylibrian:
Pushed. I squashed it as well. Please review and let me know your feedback. Thanks!

dbu:
thanks. please ping us if you need a stable release tag for this. we just did one this morning, i would think to let things sit a bit until we see if you find other bugfixes that we should include in the release.

andylibrian:
No problem, not in a rush. It can wait for the next stable release. Thanks again.

Issue #267, opened 2015-05-02
implement CAST and handle ISCHILDNODE joins


lsmith77:
```
SELECT u.*
             FROM [nt:unstructured] AS u
              INNER JOIN [nt:unstructured] AS c ON ISCHILDNODE(c, u)
             WHERE ISDESCENDANTNODE(u, [$newsDataPath])
               AND (u.publishable = true)
               AND (u.publishStartDate IS NULL OR u.publishStartDate <= CAST('$now' AS DATE))
               AND (u.publishEndDate IS NULL OR u.publishEndDate > CAST('$now' AS DATE) OR u.publishEndDate = CAST('2008-01-01T00:00:00.000+01:00' AS DATE))
               AND u.[phpcr:class] = 'Liipch\\NewsBundle\\Document\\NewsItem'
               AND NAME(c) = 'phpcr_locale:$locale'
             ORDER BY u.publishStartDate DESC
```

lsmith77:
for the ISCHILDNODE join some inspiration can maybe be taken from https://github.com/jackalope/jackalope-doctrine-dbal/pull/178

PR #311, opened 2016-01-11
update tests and skip date query test until implemented

dbu:
testing https://github.com/phpcr/phpcr-api-tests/pull/176

CAST is not supported yet, see also #267. note that i tried without cast, just with the string, and that did not work either.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.11",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.11",
+        "phpcr/phpcr-api-tests": "dev-test-date as 2.1.13",
```
lsmith77: https://github.com/phpcr/phpcr-api-tests/releases/tag/2.1.13

PR #315, opened 2016-01-22
drop 5.3.3 build as it no longer works on travis

dbu:


dbu:
@lsmith if you agree, i will do the same at jackalope-jackrabbit.

i propose that the next minor version should drop 5.3 support alltogether.

dbu:
@lsmith77 if you agree, i will do the same at jackalope-jackrabbit.

i propose that the next minor version should drop 5.3 support alltogether.

(and sorry for the ping, luke `@lsmith`)

lsmith77:
fine for me

PR #318, opened 2016-02-11
Add the possibility of using nested joins

richard-fizz:
Adds the posibility to use a 'join join' on the left side of the joined selector

Depends on
- https://github.com/phpcr/phpcr-utils/pull/160
- https://github.com/phpcr/phpcr-api-tests/pull/177

PR #319, opened 2016-02-12
Add support for nested JOIN queries

dbu:
cleaned up version of #318

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -124,7 +124,7 @@ class QOMWalker
     }
 
     /**
-     * @param QOM\QueryObjectModelInterface $qom
+     * @param QueryObjectModel $qom
      *
      * @return string
      */
@@ -189,7 +189,9 @@ class QOMWalker
      * Validates the nodeTypes in given source
      *
      * @param QOM\SourceInterface $source
+     *
      * @return QOM\SelectorInterface[]
+     *
      * @throws \PHPCR\Query\InvalidQueryException
      */
     protected function validateSource(QOM\SourceInterface $source)
@@ -208,6 +210,7 @@ class QOMWalker
 
     /**
      * @param QOM\SelectorInterface $source
+     *
      * @throws \PHPCR\Query\InvalidQueryException
      */
     protected function validateSelectorSource(QOM\SelectorInterface $source)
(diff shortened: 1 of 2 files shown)


src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -288,22 +293,52 @@ public function walkSelectorSource(QOM\SelectorInterface $source)
...
+     * @throws \BadMethodCallException if the provided JoinCondition has no valid way of getting the right selector
+     */
+    private function getRightJoinSelector(QOM\JoinConditionInterface $right)
+    {
+        if ($right instanceof QOM\ChildNodeJoinConditionInterface) {
+            return $right->getParentSelectorName();
+        } elseif ($right instanceof QOM\DescendantNodeJoinConditionInterface) {
+            return $right->getAncestorSelectorName();
+        } elseif (method_exists($right, 'getSelector2Name')) {
+            return $right->getSelector2Name();
```
dbu: @dantleech does this make sense? is there a better solution that works without method_exists?

richard-fizz: Could use an interface that implements getSelector2Name, but that would mean all join interfaces that implement getSelector2Name would have to extend that new interface.

dbu: okay, we can't just add new interfaces for this. but you could do `$right instanceof QOM\SameNodeJoinConditionInterface || $right instanceof QOM\EquiJoinConditionInterface`. those are the only qom interfaces with the getSelector2Name. i would prefer that, the method_exists is probably slower and feels strange. the PHPCR interfaces extending QOM\JoinConditionInterface are those 4. to hit the error situation, somebody would need to write their own QOM class implementing JoinConditionInterface directly instead of any of the 4 specific interfaces.

richard-fizz: I've changed this in my fork https://github.com/richard-fizz/jackalope-doctrine-dbal
Is there a way I can push this to this PR?

dantleech: @richard-fizz one "easy" way would be to

- create a new local branch based on this branch (`$ git branch my_new_nested_join_branch nested-join`)
- checkout that branch
- apply your change
- commit, push your changes to your account
- create a new pull request against THIS branch...

dbu:
i simply added the change to the squashed commits so its already done :-)

i am waiting for travis-ci to become green, then will merge. thanks a lot for the contribution @richard-fizz !

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
richard-fizz: @dantleech Thanks for the general Git knowledge. Might refer to this later for similar issues :)

richard-fizz:
Thanks!

dbu:
released in 1.2.6

PR #321, opened 2016-03-24
Replace all spaces with underscores in cache keys.

wecodenlsystem:
When using MemcachedCache, the CachedClient can not save results to the cache, because the keys contain whitespace.
This change adds a 'sanitizer' function that replaces the spaces with underscores. This function can be changed  when using another cache backend that has other key requirements.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -26,12 +26,25 @@ class CachedClient extends Client
      */
     private $caches;
 
+    /**
+     * @var \Closure that accepts a cache key as argument and returns it after sanitizing it.
+     */
+    private $keySanitizer;
+
     public function __construct(FactoryInterface $factory, Connection $conn, array $caches = array())
     {
         parent::__construct($factory, $conn);
 
         $caches['meta'] = isset($caches['meta']) ? $caches['meta'] : new ArrayCache();
         $this->caches = $caches;
+        $this->keySanitizer = function($cacheKey){ return str_replace(' ', '_', $cacheKey);};
+    }
+
+    /**
+     * @param \Closure $sanitizer
+     */
+    public function setKeySanitizer(\Closure $sanitizer){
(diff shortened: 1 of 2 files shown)


src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -26,12 +26,26 @@ class CachedClient extends Client
...
+     */
+    private $keySanitizer;
+
     public function __construct(FactoryInterface $factory, Connection $conn, array $caches = array())
     {
         parent::__construct($factory, $conn);
 
         $caches['meta'] = isset($caches['meta']) ? $caches['meta'] : new ArrayCache();
         $this->caches = $caches;
+        $this->keySanitizer = function ($cacheKey) { return str_replace(' ', '_', $cacheKey);};
```
dbu: i was a bit worried about BC breaks, but even if somebody does deploy without resetting the cache, they will just get cache misses, nothing worse.

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -7,9 +7,13 @@
 
 class CachedClientTest extends FunctionalTestCase
 {
+    /** @var \PHPUnit_Framework_MockObject_MockObject */
+    private $cacheMock;
+
     protected function getClient(Connection $conn)
     {
-        return new CachedClient(new \Jackalope\Factory(), $conn);
+        $this->cacheMock = $this->getMock('\Doctrine\Common\Cache\MemcachedCache');
```
dbu: it seems this class is not always available on travis-ci. can we mock the interface instead of a concrete implementation? or base the mock on the ArrayCache that is the default cache used in the client if none is specified.

wecodenlsystem: I have changed it to ArrayCache.

dbu:
thanks a lot!

dbu:
released in 1.2.6

PR #323, opened 2016-03-29
Invalid state when node is deleted in a different session.

dantleech:
We have encountered problems with the new DoctrinePHPCRBundle when it shuts down the kernel. It calls "clean" on the document manager and then logs out of the session(s).

The problem is that calling "clean" invokes a `refresh(false)` and in our test cases it may be that some of the nodes managed in the current kernel have been deleted from storage in another, so our system explodes with:

```
PHPCR\InvalidItemStateException: Item /article is deleted
```

when the kernel is shutdown.

I think that in this case nodes should be de-registered (and so accessing them would throw a "node deleted" error) - 

if you `refresh($keepChanges = false)` and this is the case, it means that you may encounter this error later rather than sooner (mostly never) - not sure if that is better or worse tbh.

We could also address this in the Bundle somehow.

### DIFF
--- a/tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
+++ b/tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
@@ -30,6 +30,29 @@ class ClientTest extends FunctionalTestCase
         $this->assertEquals(1, count($result->getNodes()));
     }
 
+    /**
+     * If the node was deleted in another session and we refresh the session
+     * what should happen?
+     */
+    public function testRefreshWithDeletedNodes()
+    {
+        $root = $this->session->getNode('/');
+        $article = $root->addNode('article');
+        $article->setProperty('foo', 'bar');
+        $article->setProperty('bar', 'baz');
+        $article->setProperty('bar', 'baz');
+        $this->session->save();
+
+        // simulate node being deleted in another session.
+        $nbRows = $this->getConnection()->exec('DELETE FROM phpcr_nodes WHERE path="/article"');
+        $this->assertEquals(1, $nbRows);
+
+        $this->session->refresh(false);
+
+        // throws PathNotFoundException
+        $article->getProperty('foo');
+    }
+
     public function testAddNodeTypes()
     {
         $workspace = $this->session->getWorkspace();


dantleech:
Looking at the Doctrine ORM, it seems if I perstist an Entity, delete it directy from the database, then refresh it - nothing happens and it is still "managed" If I then re-persist it, it is not persisted. Strange ..

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
```diff
@@ -30,6 +30,29 @@ public function testQueryNodes()
...
+        $this->session->save();
+
+        // simulate node being deleted in another session.
+        $nbRows = $this->getConnection()->exec('DELETE FROM phpcr_nodes WHERE path="/article"');
+        $this->assertEquals(1, $nbRows);
+
+        $this->session->refresh(false);
+
+        // throws PathNotFoundException
+        $article->getProperty('foo');
```
dbu: this is expected behaviour. the node was deleted so accessing the property should throw an exception.

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
```diff
@@ -30,6 +30,29 @@ public function testQueryNodes()
...
+        $article->setProperty('foo', 'bar');
+        $article->setProperty('bar', 'baz');
+        $article->setProperty('bar', 'baz');
+        $this->session->save();
+
+        // simulate node being deleted in another session.
+        $nbRows = $this->getConnection()->exec('DELETE FROM phpcr_nodes WHERE path="/article"');
+        $this->assertEquals(1, $nbRows);
+
+        $this->session->refresh(false);
```
dbu: this should not throw execptions when a node is deleted.

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
```diff
@@ -30,6 +30,29 @@ public function testQueryNodes()
...
+    {
+        $root = $this->session->getNode('/');
+        $article = $root->addNode('article');
+        $article->setProperty('foo', 'bar');
+        $article->setProperty('bar', 'baz');
+        $article->setProperty('bar', 'baz');
+        $this->session->save();
+
+        // simulate node being deleted in another session.
+        $nbRows = $this->getConnection()->exec('DELETE FROM phpcr_nodes WHERE path="/article"');
```
dbu: can we harden this query by adding the workspace? 

or even better, there are tests in phpcr-api-tests about concurrency using two sessions, maybe you could use that to not bypass the abstractions. then you could move that test to phpcr-api-tests. its an all api behaviour.

there is CombinedManipulationsTest::testRemoveOtherSessionRefreshKeepChanges but it seems no equivalent test for removing in a different session and then refresh discarding changes... looking at CombinedManipulationsTest i feel we should split out a ConcurrencyTest class for all thing with 2 sessions, and probably a RefreshTest for all those refresh things... but i am digressing.

can you do a PR to add the test there, with PRs on both jackalope to see what happens?

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
dantleech: indeed, I was just checking what would happen.

tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
dantleech: ok, thats the issue then. I will make it so.

PR #324, opened 2016-03-31
test php 7

dbu:


dantleech:
Strange - the build is timing out on PHP7.

dbu:
yep. looks to me like something is going into a deadlock. would need to
try with phpunit -tap to see which test blocks, and then look if there
is something specific about that test / try skipping it for php 7 to see
if its the only one.

not a high priority for me right now, just wanted to see where we stand
on php7. happy if somebody who needs this to run can continue...

vondrasekpavel:
Hi, 

I tried it and it fails on CndParserTest in PhpcrUtils.

On localhost I temporary fixed it with this:

snimek obrazovky 2017-01-18 v 11 53 30

dbu:
pavel found that the parser goes into an endless loop on:

```php
$cndParser = new CndParser($session->getWorkspace()->getNodeTypeManager());

$cnd = <<<EOT
la /*
EOT;

$cndParser->parseString($cnd);
```

there is a bunch of tests above [that test](https://github.com/phpcr/phpcr-api-tests/blob/master/tests/PhpcrUtils/CndParserTest.php#L110), so it seems an issue with handling this particular error.

dbu:
fixed in #338 (apart from the parser error https://github.com/phpcr/phpcr-utils/issues/168, which is skipped for now)

PR #326, opened 2016-03-31
Do not change TZ of \DateTime assoc. with Node

dantleech:
The current code "normalizes to UTC for storage and uses the system timezone on hydration" according to this PR:

https://github.com/jackalope/jackalope-doctrine-dbal/pull/109

But this means that the \DateTime object we have before `$session->save()` is not equivilent to the date time object we have after, which is inconsistent.

This PR clones the \DateTime object before normalizing it to UTC for storage.


dbu:
ups. the test could actually go into phpcr-api-tests - we might have a similar problem with jackrabbit. if you could add a test there that would be great.

PR #327, opened 2016-04-04
Use multiple field comparator for sqlite and other DBs

dantleech:
Only MySQL currently supports comparison of multiple value fields with numerical properties.

This PR applies the same logic that we have for strings for numbers in other implementations other than MySQL.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.14",
+        "phpcr/phpcr-api-tests": "2.1.15",
         "phpunit/phpunit": "4.7.*",
         "phpunit/dbunit": "~1.3"
     },
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -690,6 +690,10 @@ class QOMWalker
             );
         }
 
+        if ('=' === $operator) {
+            return $this->sqlXpathComparePropertyValue($alias, $property, $literalOperand->getLiteralValue(), $operator);
+        }
+
         return sprintf(
             '%s %s %s',
             $this->sqlXpathExtractNumValue($alias, $property),


src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -690,12 +690,7 @@ public function walkNumComparisonConstraint(QOM\PropertyValueInterface $property
             );
         }
 
-        return sprintf(
-            '%s %s %s',
-            $this->sqlXpathExtractNumValue($alias, $property),
-            $operator,
-            $literalOperand->getLiteralValue()
-        );
+        return $this->sqlXpathComparePropertyValue($alias, $property, $literalOperand->getLiteralValue(), $operator);
```
dantleech: I actually wonder if we shouldn't do this for all implemenations anyway - as that method handles mysql, postgres and sqlite. The logic for MySQL is however different.

dantleech: Also not sure how this is working with other impls. as the logic this replaces casts the integer. Perhaps this should be done only for Sqlite

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -690,6 +690,10 @@ public function walkNumComparisonConstraint(QOM\PropertyValueInterface $property
             );
         }
 
+        if ('=' === $operator) {
+            return $this->sqlXpathComparePropertyValue($alias, $property, $literalOperand->getLiteralValue(), $operator);
```
dantleech: I think this should work in all `=` cases?

composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require-dev": {
         "psr/log": "~1.0",
-        "phpcr/phpcr-api-tests": "2.1.14",
+        "phpcr/phpcr-api-tests": "dev-multi_num_comp as 2.1.15",
```
dbu: change to a tag before merging

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
dbu: i hope so. the tests don't reveil any issues so i guess its right.

dantleech:
Thanks David

dbu:
well, thanks for hunting down the bug, dan!

webmozart:
Any ETA for a release including this fix?

webmozart:
Ah, apparently it's in 1.2.7 already!

Issue #305, opened 2015-10-20
ISDESCENDANTNODE does not work with path containing a dot


danrot:
Somebody has encountered an issue in combination with Sulu: https://github.com/sulu-io/sulu-standard/issues/536#issuecomment-149315457

I could track down the problem to the fact that this `ISDESCENDANTNODE` function does not work correctly, when the path contains a dot. So the following query returns nothing, although the node at the given path has child nodes:

```
SELECT * FROM [nt:unstructured] AS page WHERE ISDESCENDANTNODE(page, '/cmf/htl.do/contents');
```

I guess this should be working, since the dot doesn't have any special meaning (at least none that I am aware of), and this query works like a charm when used with jackrabbit.

dbu:
yes, this indeed should work. it looks like we have some problem with
escaping here. can you find out what the sql query is that gets executed
for this?

danrot:
Not sure what the best way to do this is... Tried to find it in the logs, the only thing that made sense was this one:

```
[2015-11-02 23:06:12] doctrine.DEBUG:                SELECT * FROM phpcr_nodes               WHERE path = :path                 AND workspace_name = :workspace               ORDER BY depth, sort_order ASC {":path":"/cmf/sulu.io",":pathd":"/cmf/sulu.io/%",":workspace":"default"} []
```

dbu:
> Not sure what the best way to do this is...

try to work with the phpcr-api-tests. there should be a test about
ISDESCENDANTNODE i would hope, that you could change to a path with dot,
and if we miss the whole test, you could add one.

dbu:
btw, could this be related to https://github.com/phpcr/phpcr-utils/pull/104 ?

i never wrapped this up. there was some weird regression that i did not figure out. maybe check if that is the reason for this problem and if so, look into it.

danrot:
Seems like I have messed something up back then... I just realized that this is an error in Sulu. Should be fixed with https://github.com/sulu/sulu/pull/3039, so this issue can be closed.

danrot:
I am afraid I was too fast, mixed that up with another issue we've had 🙈
Sorry, was a tough week.

dbu:
is this because of https://github.com/phpcr/phpcr-utils/pull/104 or something else?

danrot:
I don't know... Tried to test it, but when I check out the other branch and execute the query I get the following error:
```
  [Symfony\Component\Debug\Exception\FatalThrowableError]
[...]

Issue #248, opened 2015-03-03
Path is limited to 255 chars


dantleech:
What happens when somebody creates a path which is longer than 255 chars?

dbu:
write a test and find out :-P

i would expect at least mysql to just cut off after 255 chars. and indeed this sounds like a potential pitfall. at the very least we should throw a clear exception. not sure about performance implications if we take a higher value. oh, and seems newer versions of mysql can do longer varchar: http://stackoverflow.com/questions/13506832/what-is-the-mysql-varchar-max-size so maybe we should just make that configurable and default to a higher value. 
no idea what the cost is (in terms of storage or read / write performance) of putting it to say 10k.

reading this, we should just use a TEXT field in postgres: http://www.postgresql.org/docs/8.3/static/datatype-character.html

for sqlite, it seems to not matter at all what we do - it already handles any reasonable path and quite a bit more, if this answer is correct: http://stackoverflow.com/questions/6109532/what-is-the-maximum-size-limit-of-varchar-data-type-in-sqlite

lsmith77:
I think the issue is a limitation in MySQL and indexes. not sure if this limitation still exists but yeah, we should not make other RDBMS suffer ..

dantleech:
For the record this came up in IRC when somebody was getting "the path already exists" errors, I assumed it was related to the fact that his path was 320 characters long.

dbu:
sounds plausible, yes.

vondrasekpavel:
Hi,

I'm working on this issue and I examined how to enlarge varchar in MySQL with index.

I read that in MySQL v.5.5.14+ you can do this on varchar(1024):
SET @@global.innodb_large_prefix = 1
ALTER TABLE `phpcr_nodes` ADD INDEX(`path`)

global.innodb_large_prefix allows index size 3072.

In Postgres I read that text is same like varcher and there are no performance differences between these two types. I tried to create table with column, added index on it and all worked.

In SQLite I think that varchar size is unlimited: https://sqlite.org/faq.html#q9

Do you think that is better to use biggest available path size on database platform or be consistent between platforms? 

I think we can do:

1) add option for MySQL to create/update schema with varchar(1024) if it technically possible; on Postgres use text and on SQLite text too (I think that varchar(n) is same like text) and use available path size.

or

[...]

Issue #246, opened 2015-02-16
ISCHILDNODE and ISDESCENDANTNODE not working with %(wildcard)


oste:
I am trying to leverage ```ISCHILDNODE``` with a % wildcard using the Query Builder. I am constructing my query like this
```
$qb->from()->document('AppBundle\Document\ContentDocument', 'd');
//$qb->where()->descendant('/my/%/posts', 'd');
$qb->where()->child('/my/%/posts', 'd');
```
This gives me a SQL-2 query that looks like this ```SELECT * FROM [nt:unstructured] AS d WHERE (ISCHILDNODE(d, [/my/%/posts])``` Unfortunately this is not working :(

It is worth noting that ISDESCENDANTNODE works as expected ```SELECT * FROM [nt:unstructured] AS d WHERE (ISDESCENDANTNODE(d, [/my/%/posts])```

@dantleech was able to confirm that neither of these work with Jackrabbit in #symfony-cmf irc chat. However they should

dbu:
this sounds strange. a first step to solve this would be to add such queries to https://github.com/phpcr/phpcr-api-tests/blob/master/tests/06_Query/QuerySql2OperationsTest.php and do a pull request with those. then we can investigate what actually goes on. do you want to do that @oste ?

oste:
Thanks @dbu, I will try to find some time this week to get that pull request started.

Issue #217, opened 2014-08-25
[query] Selector does not allow specification of mixin


dantleech:
The doctrine dbal selector does not seem to support selecting from mixin types, for example `sulu:content` is a mixin type:

```
PHPCRSH > SELECT * FROM [sulu:content] AS c WHERE ISCHILDNODE('/cmf/sulu_io/contents/test') AND c.[i18n:de-state] = 2;
+------+
| Path |
+------+
```

```
PHPCRSH > SELECT [jcr:mixinTypes] FROM [nt:unstructured] AS c WHERE ISCHILDNODE('/cmf/sulu_io/contents/test') AND c.[i18n:de-state] = 2;
+-----------------------------------+------------------+
| Path                              | c.jcr:mixinTypes |
+-----------------------------------+------------------+
| /cmf/sulu_io/contents/test/foobar | [0] sulu:content |
| /cmf/sulu_io/contents/test/barfoo | [0] sulu:content |
+-----------------------------------+------------------+
```

http://www.day.com/specs/jcr/2.0/6_Query.html#6.7.3 Selector:

> the node’s primary node type is nodeType, or
> the node’s primary node type is a subtype of nodeType, or
> the node has a mixin node type that is nodeType, or
> the node has a mixin node type that is a subtype of nodeType

Jackrabbit supports mixin as selector.

PR #334, opened 2017-01-18
prefer update instead of insert in registerNodeTypes

vondrasekpavel:
Hi,

I don't like that if I call registerNodeTypes method that it removes node type and creates it with new id. I think that it isn't update :)

I think that use update is prettier :)

Thank you :)

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -1998,8 +1998,6 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     /**
      * Fetch a user-defined node-type definition.
      *
-     * @param string $name
-     *
      * @return array
      */
     protected function fetchUserNodeTypes()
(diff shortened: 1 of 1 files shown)


vondrasekpavel:
I don't know what is wrong with PHP 5.6 but I didn't update dependencies :(

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2084,27 +2082,36 @@ public function registerNodeTypes($types, $allowUpdate)
                 $query = "SELECT * FROM phpcr_type_nodes WHERE name = ?";
                 $result = $this->getConnection()->fetchColumn($query, array($type->getName()));
                 if ($result) {
-                    $this->getConnection()->delete('phpcr_type_nodes', array('node_type_id' => $result));
                     $this->getConnection()->delete('phpcr_type_props', array('node_type_id' => $result));
                     $this->getConnection()->delete('phpcr_type_childs', array('node_type_id' => $result));
```
dbu: so we delete all props and children mappings for the type, but not the type entry itself. this could lead us to an inconsistent situation when updating fails for some reason. but i don't see a better solution

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -2084,27 +2082,36 @@ public function registerNodeTypes($types, $allowUpdate)
...
-                    'is_mixin' => $type->isMixin() ? 1 : 0,
-                    'queryable' => $type->isQueryable() ? 1 : 0,
-                    'orderable_child_nodes' => $type->hasOrderableChildNodes() ? 1 : 0,
-                    'primary_item' => $type->getPrimaryItemName(),
-                ));
+                if ($allowUpdate && $result) {
+                    $this->getConnection()->update('phpcr_type_nodes', $data, array('node_type_id' => $result));
+                } else {
+                    $this->getConnection()->insert('phpcr_type_nodes', $data);
+                }
```
dbu: while we clean this up, why not do

```php
if ($result) {
    if (!$allowUpdate) {
        throw ...
    }
    $this->getConnection()->update...
} else {
   $this->getConnection()->insert...
}
```

and get rid of the try-catch. as we know if the result exists or not, we can handle it this way, and not hide internal errors with missing validation and invalid queries as a node type exists..

vondrasekpavel: What do you think about it now?

I don't like this because it queries database in cycle :(
snimek obrazovky 2017-01-18 v 14 31 35

dbu:
cool! i think this is indeed quite a bit better.

> I don't like this because it queries database in cycle :(

agreed, that is not as resource efficient as could be. then again, node type registration is not something that is expected to happen very often. at most during the deployment, if no check is done to see if types need updating.

i think i will merge this as is. if you want to further tweak this, its easier to do in a separate pull request.

re the build failures: i just restarted [a 5.6 build that worked a month ago](https://travis-ci.org/jackalope/jackalope-doctrine-dbal/jobs/186113273) and now it fails. so definitely not related to this PR, but some regression with the doctrine libraries dependency resolution...

vondrasekpavel:
Thank you for your merge. :)

Can you give me some recommendations which parts of PHPCR / Jackalope / Jacklope Dbal need repair or improve? I want help you but I don't know where to start.

dbu:
i love that question! i looked through open issues and if you want to dig through some issues with querying, there are #305, #267, #246 and #217. #248 would be something we want to solve someday, but will need some investigating to not end up with a huge performance regression.

on core jacaklope, this would be an interesting improvement https://github.com/jackalope/jackalope/pull/287 - we did not find time to wrap it up. this one sounds like an easily fixable issue: https://github.com/jackalope/jackalope/issues/314

PR #337, opened 2017-01-27
change collation in query to adapt to charset

danrot:
This change is related to https://github.com/doctrine/DoctrinePHPCRBundle/pull/270. If a connection to the [RepositorySchema](https://github.com/jackalope/jackalope-doctrine-dbal/blob/75bfb54453f224a9e848cee70cc85bf8d48c4556/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php#L37) with a charset of `utf8mb4` is passed it is not allowed to use `utf8_bin` in a select statement.

I simply changed it in here to `utf8mb4_bin`, which will work then. However, I know that this can't be the final solution. The question is how that one should look like. We could use the `connection` to get the charset and use an if in here, or use the charset and simply postfix it with `_bin`, maybe check before if this collation really exists (is that somehow possible? What should be done in case it isn't?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+1.3.0
+-----
+
+* Added auto-detection of binary collation for MySQL, so that not always utf8_bin is used.
+  A new `setCaseSensitiveEncoding` method has been introduced, which can be used to override
+  the auto detected value.
+
 1.2
 ---
 
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -206,6 +206,11 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     private $nodeProcessor;
 
     /**
+     * @var string
+     */
+    private $caseSensitiveEncoding;
+
+    /**
      * @param FactoryInterface $factory
      * @param Connection       $conn
      */
@@ -437,6 +442,34 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     public function setCheckLoginOnServer($bool)
     {
         $this->checkLoginOnServer = $bool;
+    }
+
+    /**
+     * This will control the collate which is being used on MySQL when querying nodes. It will be autodetected by just
+     * appending _bin to the current charset, which is good enough in most cases.
+     *
+     * @param string $encoding
+     */
+    public function setCaseSensitiveEncoding($encoding)
+    {
+        $this->caseSensitiveEncoding = $encoding;
+    }
+
+    /**
+     * Returns the collate which is being used on MySQL when querying nodes.
+     *
+     * @return string
+     */
+    private function getCaseSensitiveEncoding()
+    {
+        if (!$this->caseSensitiveEncoding) {
+            $params = $this->conn->getParams();
+            $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
+
+            return $this->caseSensitiveEncoding = $charset === 'binary' ? $charset : $charset . '_bin';
+        }
+
+        return $this->caseSensitiveEncoding;
     }
 
     protected function workspaceExists($workspaceName)
@@ -1540,7 +1573,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
             $query = 'SELECT id FROM phpcr_nodes WHERE identifier = ? AND workspace_name = ?';
         } else {
             if ($this->getConnection()->getDriver() instanceof \Doctrine\DBAL\Driver\PDOMySql\Driver) {
-                $query = 'SELECT id FROM phpcr_nodes WHERE path COLLATE utf8_bin = ? AND workspace_name = ?';
+                $query = 'SELECT id FROM phpcr_nodes WHERE path COLLATE ' . $this->getCaseSensitiveEncoding() .' = ? AND workspace_name = ?';
             } else {
                 $query = 'SELECT id FROM phpcr_nodes WHERE path = ? AND workspace_name = ?';
             }


dbu:
where can do you configure the encoding? would it be possible to have that place also configure a binary encoding to go with that, which by default is just $encoding.'_bin' but can be customized, then read that value in the client for that query?

or alternatively, have the client check if the driver is on utf8 and if not not use any collate in the query? from what i understand, collate is only used as a performance improvement. or is it also about being case sensitive or a vs ä?

danrot:
I've been looking for the commit in which this change is, and I've found it: https://github.com/jackalope/jackalope-doctrine-dbal/commit/5273f180df5c3f7ecca4b3da81ba5edfd864e3b0

Looks like it is about case-sensitivity. So I guess it it is not possible to simply remove it in case of UTF-8, right? Seems there was a problem at some point.

The current configuration is from the DoctrineBundle, so not really related to Jackalope. If we want to introduce a new configuration we would have to find a place for it first.

I have just checked the available collations on my system, and there is a `*_bin`-collation for every character set, with `binary` being the only exception. So should we maybe take the "adding _bin"  option and just check for the binary character set?

dbu:
i guess that is about as save as the current code using hardcoded utf8_bin. maybe we can add a setCaseSensitiveEncoding method on the client class, as a last resort for people that have a special case with some requirements? something like:

```
private $caseSensitiveEncoding;
private function getCaseSensitiveEncoding()
{
 if (!$this->caseSensitiveEncoding) {
   $this->caseSensitiveEncoding = get driver collation .'_bin';
 }

 return $this->caseSensitiveEncoding;
}
```

(and only call this in the mysql case.)

danrot:
@dbu But that's only useful if it is at least protected, right? If we make it private nobody can change it, so there would be no reason to introduce that. Or am I overseeing something now? 🙈

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1539,8 +1539,13 @@ private function getSystemIdForNode($identifier, $workspaceName = null)
         if (UUIDHelper::isUUID($identifier)) {
             $query = 'SELECT id FROM phpcr_nodes WHERE identifier = ? AND workspace_name = ?';
         } else {
-            if ($this->getConnection()->getDriver() instanceof \Doctrine\DBAL\Driver\PDOMySql\Driver) {
-                $query = 'SELECT id FROM phpcr_nodes WHERE path COLLATE utf8_bin = ? AND workspace_name = ?';
+            $params = $this->conn->getParams();
+            $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
```
danrot: There's a problem with that solution... The `charset` option and the default charset with which a table is created can be different... How should we handle that? Couldn't find another way to get that from the `SchemaManager` in Doctrine for the specific table 😕 

Just to be clear, I am talking about this situation in the config:

```yaml
doctrine:
    dbal:
        charset: utf8mb4
        default_table_options:
            charset: utf8mb4
            collate: utf8mb4_unicode_ci
```

dbu: oh well. but in that case lets expect people to set the case sensitive encoding manually?

i think we really must avoid doing a database call in every request to know this, so it must come from the connection / driver if thats possible at all. worst case we just add this new method and have a compiler pass in the phcpr-bundle that reads the dbal configuration and does something sensible?

@lsmith77 can you help us here to read the right information?

dbu:
> @dbu <https://github.com/dbu> But that's only useful if it is at least
> protected, right? If we make it private nobody can change it, so there
> would be no reason to introduce that. Or am I overseeing something now? 🙈

i missed to write the setter in my illustration...

i would have a

/**
 * Set table encoding to use for case sensitive lookups in MySQL.
 */
public function setCaseSensitiveEncoding($encoding)

src/Jackalope/Transport/DoctrineDBAL/Client.php
lsmith77: hmm would need to dig in here too .. but I agree we should do as little work as possible at run time ..
so what we can do at install time or at config compile time the better.

danrot: Just to be clear: The current state doesn't involve any database call at all, or am I missing something?

dbu: no, currently not. i was just saying that we must not end with something that does.

danrot: And you are also saying that the current solution is not satisfying, right? Would it be enough to add the `setCaseSensitiveEncoding` method?

dbu: yes, i would say so. have a setter, and a get method that does the code you added above and below, so that this code can just say if mysql and not binary, do collate with $this->getCaseSensitiveEncoding()

danrot:
@dbu Added these methods, is it ok now?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -439,6 +444,23 @@ public function setCheckLoginOnServer($bool)
...
+        $this->caseSensitiveEncoding = $encoding;
+    }
+
+    private function getCaseSensitiveEncoding()
+    {
+        if (!$this->caseSensitiveEncoding) {
+            $params = $this->conn->getParams();
+            $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
+
+            return $charset === 'binary' ? $charset : $charset . '_bin';
```
dbu: please do `$this->caseSensitiveEncoding = $charset === 'binary' ? $charset : $charset.'_bin';` instead of return, so its only done once

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -439,6 +444,23 @@ public function setCheckLoginOnServer($bool)
         $this->checkLoginOnServer = $bool;
     }
 
+    public function setCaseSensitiveEncoding($encoding)
```
dbu: please add a docblock explaining that this is to control what collate should be used when querying for nodes. and mention that its only relevant for mysql and usually the autodetection of appending _bin to the charset is good enough.

danrot:
@dbu Fixed your comments.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+1.2.8
```
dbu: if we merge this change into master, it will become 1.3, as master is for 1.3 and there are other changes in it. do you urgently need a fixed release? vondrasekpavel is busy doing lots of cleanups and fixes, we would release sometimes in february.

for an urgent fix, i would have to update the 1.2 branch to have everything that is in 1.2 (it currently looks massively outdated: https://github.com/jackalope/jackalope-doctrine-dbal/compare/1.2...1.2.8). if you need a fix urgently, i can do that and then merge your branch into 1.2 and tag 1.2.9 instead.

danrot: Ah, I thought master is the next hotfix release... Actually @alexander-schranz has to tell how fast he needs that fix, for me the next minor release is also good 😃

danrot: Have talked to @alexander-schranz, it's not that important. Changed it to 1.3

dbu:
great, then it will go into 1.3. i guess we should release that soonish, quite a few fixes and improvements going on in the last weeks! thanks daniel!

danrot:
Cool, thank you! Now only https://github.com/doctrine/DoctrinePHPCRBundle/pull/270 is missing in order for it to work in a symfony environment :-)

dbu:
thanks for the reminder, merged that as well.

PR #340, opened 2017-01-31
updated documentation

vondrasekpavel:


dbu:
thanks! that was some sloppy examples. love how we had once a leading \ for the FQN, then on the next line it starts without a backslash...

vondrasekpavel:
David, thank you for merge! I plan use constants in example but now don't work :)

PR #341, opened 2017-01-31
phpcr 2.1.3

vondrasekpavel:
We're compatible with new version I think :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^5.6|^7.0",
         "doctrine/dbal": ">=2.4.5,<3.0.x-dev",
-        "phpcr/phpcr": "~2.1.2",
+        "phpcr/phpcr": "~2.1.3",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": "^5.6|^7.0",
         "doctrine/dbal": ">=2.4.5,<3.0.x-dev",
-        "phpcr/phpcr": "~2.1.2",
+        "phpcr/phpcr": "~2.1.3",
```
dbu: this means we are not compatible with 2.1.2 anymore. i think there is no need to increase the minimal version. `~` means at least this version, or any newer patch release of 2.1.

vondrasekpavel: Ok, you're right will close it. :)

Today I'm confused because I thought that I updated this constants: https://github.com/phpcr/phpcr/blob/master/src/PHPCR/PropertyInterface.php :o

PR #342, opened 2017-02-27
user node types optimization

vondrasekpavel:
I added caching user node types to avoid querying in every getNodeTypes method call.

Before:
snimek obrazovky 2017-02-26 v 22 50 29

After:
snimek obrazovky 2017-02-26 v 22 50 59

@danrot can you test it on Sulu, please? Thank you :)

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -212,6 +212,11 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     private $caseSensitiveEncoding;
 
     /**
+     * @var array
+     */
+    private $userNodeTypes;
+
+    /**
      * @param FactoryInterface $factory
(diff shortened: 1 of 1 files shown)


vondrasekpavel:
I think we should close this issue: https://github.com/jackalope/jackalope/pull/287 because built-in node types are fast on my php 7.1

dbu:
looks like a very good improvement, thanks! please move this logic into the [CachedClient](https://github.com/jackalope/jackalope-doctrine-dbal/blob/master/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php) class (simply overwriting fetchUserNodeTypes should do the trick i think).

you also need to invalidate this cache in all methods that change node types. the node type management interface only has registerNodeTypes - i think thats the only method that needs to trigger invalidation.

vondrasekpavel:
Ok, my fault I haven't cached correctly :( Sorry all! I close this PR.

PR #352, opened 2017-12-06
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertNotFalse` instead os strict comparisons `!==` with `false` keyword;
- `assertObjectHasAttribute` instead of `isset` function.


dbu:
thanks!

PR #353, opened 2018-01-27
Add support for utf8mb4

alexander-schranz:
When using the following configuration for doctrine based on [Symfony 3.4 recommandation](https://symfony.com/doc/3.4/doctrine.html#configuring-the-database):

```yaml
doctrine:
    dbal:
        driver:   "%database_driver%"
        host:     "%database_host%"
        port:     "%database_port%"
        dbname:   "%database_name%"
        user:     "%database_user%"
        password: "%database_password%"
        server_version: "%database_version%"
        charset: utf8mb4
        default_table_options:
            charset: utf8mb4
            collate: utf8mb4_unicode_ci
```

The database fail because the index key is too long. From the official docs the keys should not be longer then 191 because utf8mb4 needs 4 bytes instead of the 3 byte per character.  https://dev.mysql.com/doc/refman/5.6/en/charset-unicode-conversion.html

Tested with: MySQL 5.6.21 on OSX. To avoid a BC Break it only will set the length to 191 when utf8mb4 is used.

<details>
  <summary>
Error message which did appears before this changes:</summary>

```bash
In ToolsException.php line 39:

          Schema-Tool failed with Error 'An exception occurred while
           executing 'CREATE TABLE phpcr_namespaces (prefix VARCHAR(
          192) NOT NULL, uri VARCHAR(255) NOT NULL, PRIMARY KEY(pref
          ix)) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode
          _ci ENGINE = InnoDB':

          SQLSTATE[42000]: Syntax error or access violation: 1071 Sp
          ecified key was too long; max key length is 767 bytes' whi
          le executing DDL: CREATE TABLE phpcr_namespaces (prefix VA
          RCHAR(192) NOT NULL, uri VARCHAR(255) NOT NULL, PRIMARY KE
          Y(prefix)) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_u
          nicode_ci ENGINE = InnoDB


        In AbstractMySQLDriver.php line 115:

          An exception occurred while executing 'CREATE TABLE phpcr_
          namespaces (prefix VARCHAR(192) NOT NULL, uri VARCHAR(255)
           NOT NULL, PRIMARY KEY(prefix)) DEFAULT CHARACTER SET utf8
          mb4 COLLATE utf8mb4_unicode_ci ENGINE = InnoDB':

          SQLSTATE[42000]: Syntax error or access violation: 1071 Sp
          ecified key was too long; max key length is 767 bytes


        In PDOConnection.php line 106:

          SQLSTATE[42000]: Syntax error or access violation: 1071 Sp
          ecified key was too long; max key length is 767 bytes


        In PDOConnection.php line 104:

          SQLSTATE[42000]: Syntax error or access violation: 1071 Sp
          ecified key was too long; max key length is 767 bytes
```
</details>

### DIFF
--- a/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
+++ b/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
@@ -84,7 +84,7 @@ EOT
             return self::RETURN_CODE_NO_FORCE;
         }
 
-        $schema = new RepositorySchema;
+        $schema = new RepositorySchema([], $connection);
 
         try {
             if ($input->getOption('drop')) {
--- a/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
@@ -22,6 +22,11 @@ class RepositorySchema extends Schema
     private $connection;
 
     /**
+     * @var integer
+     */
+    private $maxIndexLength;
+
+    /**
      * @param array $options The options could be use to make the table
      *                               names configurable.
      * @param Connection $connection
@@ -66,7 +71,7 @@ class RepositorySchema extends Schema
     protected function addNamespacesTable()
     {
         $namespace = $this->createTable('phpcr_namespaces');
-        $namespace->addColumn('prefix', 'string');
+        $namespace->addColumn('prefix', 'string', ['length' => $this->getMaxIndexLength()]);
         $namespace->addColumn('uri', 'string');
         $namespace->setPrimaryKey(['prefix']);
     }
@@ -74,7 +79,7 @@ class RepositorySchema extends Schema
     protected function addWorkspacesTable()
     {
         $workspace = $this->createTable('phpcr_workspaces');
-        $workspace->addColumn('name', 'string');
+        $workspace->addColumn('name', 'string', ['length' => $this->getMaxIndexLength()]);
         $workspace->setPrimaryKey(['name']);
     }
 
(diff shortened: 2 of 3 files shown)


src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
```diff
@@ -219,4 +224,30 @@ public function reset()
...
+        return $currentMaxLength < $this->maxIndexLength ? $currentMaxLength : $this->maxIndexLength;
+    }
+
+    private function isConnectionCharsetUtf8mb4()
+    {
+        if (!$this->connection) {
+            return false;
+        }
+
+        $databaseParameters = $this->connection->getParams();
```
dbu: am i correct that this does not lead to a communication with the database? (otherwise we should be worried of the overhead)

alexander-schranz:
@dbu it will throw the same exception as currently when for example the path is longer then 255:

```bash
SQLSTATE[22001]: 1406 Data too long for column 'path' at row 5
```

src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
alexander-schranz: It only reads the [configuration](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L245-L248) which is set in the [constructor](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L206) of the dbal connection so this should not impact the performance.

alexander-schranz: The only thing I don't know is why the connection is optional and not set here: https://github.com/jackalope/jackalope-doctrine-dbal/blob/1.3.1/src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php#L87

dbu: i guess its optional so that the commands can be built at all when no connection is set up yet. and i guess in the command we did not pass the connection because it was not needed. happy if you change the command to pass the connection. is this also whats used in the phpcr bundle, or do we create the schema differently?

alexander-schranz: did now also set the connection in InitDoctrineDbalCommand. In the phpcr bundle the connection seems correctly be injected in the [service definition](https://github.com/doctrine/DoctrinePHPCRBundle/blob/c9b8bbdbee712042f536b8a30902bb8614c2411c/src/Resources/config/jackalope_doctrine_dbal.xml#L14) and building the schema for sulu did work correctly for me.

src/Jackalope/Tools/Console/Command/InitDoctrineDbalCommand.php
```diff
@@ -84,7 +84,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return self::RETURN_CODE_NO_FORCE;
         }
 
-        $schema = new RepositorySchema;
+        $schema = new RepositorySchema([], $connection);
```
dbu: this seems to have super weird side effects in the schema class :-(
can you try to figure out what we should do there instead? looks like we abuse the $connection to decide between different situations...

alexander-schranz: seems only be an unexpected call in the unit test. did fix that or are there any other sideeffects?

dbu: ah, you are right, the mock was just not expecting the call. one should not atttempt to write confusing ternary operators when half falling asleep after a conference :-D

dbu:
finished in #354

thanks a lot @alexander-schranz !

dbu:
i released 1.3.2 with this fix.

alexander-schranz:
@dbu Awesome thank you!

PR #354, opened 2018-01-29
MySQL utf8mb4 limitations

dbu:
MySQL implemented utf8 support without 4 byte characters and added the utf8mb4 encoding to work around the issue. Indexes for utf8mb4 can not be more than 191 characters.

Postgres is not affected as they implemented full utf8 from the beginning. I did not find anything about this for sqlite, but assume for now that its also not affected.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
@@ -24,7 +24,7 @@ class RepositorySchema extends Schema
     /**
      * @var integer
      */
-    private $maxIndexLength;
+    private $maxIndexLength = -1;
 
     /**
      * @param array $options The options could be use to make the table
@@ -225,19 +225,24 @@ class RepositorySchema extends Schema
         }
     }
 
-    private function getMaxIndexLength($currentMaxLength = 255)
+    private function getMaxIndexLength($currentMaxLength = null)
     {
-        if (null !== $this->maxIndexLength) {
-            return $currentMaxLength < $this->maxIndexLength ? $currentMaxLength : $this->maxIndexLength;
+        if (-1 === $this->maxIndexLength) {
+            $this->maxIndexLength = null;
+
+            if ($this->isConnectionCharsetUtf8mb4()) {
+                $this->maxIndexLength = 191;
+            }
         }
 
-        $this->maxIndexLength = 255;
-
-        if ($this->isConnectionCharsetUtf8mb4()) {
-            $this->maxIndexLength = 191;
+        if ($currentMaxLength && (
+            null === $this->maxIndexLength
+            || $currentMaxLength < $this->maxIndexLength
+        )) {
+            return $currentMaxLength;
         }
-
-        return $currentMaxLength < $this->maxIndexLength ? $currentMaxLength : $this->maxIndexLength;
+        
+        return $this->maxIndexLength;
     }
 
     private function isConnectionCharsetUtf8mb4()
--- a/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
@@ -22,6 +22,11 @@ class RepositorySchema extends Schema
     private $connection;
 
     /**
+     * @var integer
+     */
+    private $maxIndexLength;
+
+    /**
      * @param array $options The options could be use to make the table
      *                               names configurable.
      * @param Connection $connection
@@ -66,7 +71,7 @@ class RepositorySchema extends Schema
     protected function addNamespacesTable()
     {
         $namespace = $this->createTable('phpcr_namespaces');
-        $namespace->addColumn('prefix', 'string');
+        $namespace->addColumn('prefix', 'string', ['length' => $this->getMaxIndexLength()]);
         $namespace->addColumn('uri', 'string');
         $namespace->setPrimaryKey(['prefix']);
     }
@@ -74,7 +79,7 @@ class RepositorySchema extends Schema
     protected function addWorkspacesTable()
     {
         $workspace = $this->createTable('phpcr_workspaces');
-        $workspace->addColumn('name', 'string');
+        $workspace->addColumn('name', 'string', ['length' => $this->getMaxIndexLength()]);
         $workspace->setPrimaryKey(['name']);
     }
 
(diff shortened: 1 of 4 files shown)


src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php
```diff
@@ -219,4 +224,35 @@ public function reset()
             $this->connection->exec($sql);
         }
     }
+
+    private function getMaxIndexLength($currentMaxLength = null)
+    {
+        if (-1 === $this->maxIndexLength) {
+            $this->maxIndexLength = null;
```
dbu: @alexander-schranz i continued on your pull request. before, we did not pass a length - i feel its better to keep it that way by default. this code is not only for mysql but also postgres and sqlite, i am afraid of side effects if we suddenly limit to 255 characters.

dbu:
@alexander-schranz while reading up on the topic, i discovered that it would also be possible to build prefix indexes that only use the first X characters of a field. i wonder if this is what we should do for the path and parent fields? paths can easily become too long.

there is of course a hefty performance penalty when fields are longer than the prefix, as those need to then be scanned without index - but the limitation on full path length seems quite arbitrary and 191 is not a lot of characters...

alexander-schranz:
@dbu I did see that with the prefixed index but don't know if this is possible in doctrine?
As you said its only a problem with mysql we should check for the database type:

The best way would be check over `DatabasePlatform` reading to the code I'm not sure if there is an additional request done:

```php
if ('mysql' !== $this->connection->getDatabasePlatform()->getName()) {
    return false;
}
``` 

Another way would be over the driver but not sure if there are any other mysql drivers:

```php
if ('pdo_mysql' !== $this->connection->getDriver()->getName()) {
    return false;
}
```

dbu:
the prefix index thing should be possible with dbal, but while checking, i realize that we also add a unique index, and that seems the killer for this idea... so i guess we leave at that.

i don't think we need to check the platform - utf8mb4 is a mysql specific invention that postgres and sqlite don't even know about, so you can't accidentally set it for those.

do you agree with merging as is? can you please check the commit i added, do you see any flaw with how i make the length `null` instead of 255 for non utf8mb4? (i checked that the column class should be happy with setting the length to null explicitly)

alexander-schranz:
@dbu thank you looks good for me and as the utf8mb4 seems to be a mysql only thing I agree to merge it.

alexander-schranz:
@dbu I tried to add utf8mb4 to the tests but could not get it to work as the tables seems to be created still with utf8 not utf8mb4 in the test setup I checked in my information_schema that there is the default set correctly for the `phpcr_tests` table. 

<details>
<summary>My tried changes</summary>

```
diff --git a/phpunit.xml.dist b/phpunit.xml.dist
index d47021c..30680d1 100644
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -18,12 +18,16 @@
 
         <var name="data_caches" value="meta,nodes" />
         <!-- Doctrine DBAL database configuration -->
-        <var name="phpcr.doctrine.dbal.driver" value="pdo_sqlite" />
-        <var name="phpcr.doctrine.dbal.path" value="phpcr_tests.db" />
+        <var name="phpcr.doctrine.dbal.driver" value="pdo_mysql" />
         <var name="phpcr.doctrine.dbal.host" value="localhost"/>
-        <var name="phpcr.doctrine.dbal.username" value="jackalope"/>
+        <var name="phpcr.doctrine.dbal.username" value="root"/>
         <var name="phpcr.doctrine.dbal.password" value=""/>
         <var name="phpcr.doctrine.dbal.dbname" value="phpcr_tests"/>
+        <var name="phpcr.doctrine.dbal.charset" value="utf8mb4"/>
+        <var name="phpcr.doctrine.dbal.default_table_options.charset">
+            <var name="charset" value="utf8mb4"/>
+            <var name="collate" value="utf8mb4_unicode_ci"/>
+        </var>
     </php>
 
     <testsuites>
diff --git a/src/Jackalope/Transport/DoctrineDBAL/Client.php b/src/Jackalope/Transport/DoctrineDBAL/Client.php
index e28d51a..aff2f2c 100644
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -464,6 +464,12 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface,
             $params = $this->conn->getParams();
             $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
 
+            if (isset($params['default_table_options']) && $params['default_table_options']['collate']) {
+                $this->caseSensitiveEncoding = $params['default_table_options']['collate'];
+
+                return $this->caseSensitiveEncoding;
+            }
+
             return $this->caseSensitiveEncoding = $charset === 'binary' ? $charset : $charset . '_bin';
         }
 
diff --git a/tests/bootstrap.php b/tests/bootstrap.php
index dfda39d..db3647d 100644
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -38,7 +38,12 @@ $dbConn = DriverManager::getConnection([
     'port'      => @$GLOBALS['phpcr.doctrine.dbal.port'],
     'user'      => @$GLOBALS['phpcr.doctrine.dbal.username'],
     'password'  => @$GLOBALS['phpcr.doctrine.dbal.password'],
-    'dbname'    => @$GLOBALS['phpcr.doctrine.dbal.dbname']
+    'dbname'    => @$GLOBALS['phpcr.doctrine.dbal.dbname'],
+    'charset'    => @$GLOBALS['phpcr.doctrine.dbal.charset'],
+    'default_table_options'    => [
+        'charset' => 'utf8mb4',
+        'collate' => 'utf8mb4_unicode_ci',
+    ]
 ]);
 
 /** Recreate database schema */
diff --git a/tests/generate_phpunit_config.php b/tests/generate_phpunit_config.php
index f635b67..05859c6 100644
--- a/tests/generate_phpunit_config.php
+++ b/tests/generate_phpunit_config.php
@@ -15,6 +15,11 @@ $config = [
         'phpcr.doctrine.dbal.username'  => 'travis',
         'phpcr.doctrine.dbal.password'  => '',
         'phpcr.doctrine.dbal.dbname'    => 'phpcr_tests',
+        'phpcr.doctrine.dbal.charset'   => 'utf8mb4',
+        'phpcr.doctrine.dbal.default_table_options' => [
+            'charset' => 'utf8mb4',
+            'collate' => 'utf8mb4_unicode_ci',
+        ],
     ],
     'pgsql'  => [
         'phpcr.doctrine.dbal.driver'    => 'pdo_pgsql',
```

<details>

dbu:
hm, but it works in your setup? i would be ok to merge as is if you confirm it works.

alexander-schranz:
@dbu 

<details>
<summary>yes works for me correctly in sulu 👍 </summary>

bildschirmfoto 2018-01-29 um 14 27 59
</details>

PR #357, opened 2018-05-03
WIP: upgrade to dbal 2.7, use doctrine constants instead of pdo

### LINKED ISSUE #355: avoid using deprecated PDO:: constants in calls to doctrine dbal
according to the [upgrade notes](https://github.com/doctrine/dbal/blob/master/UPGRADE.md#upgrade-to-27), we should stop using the pdo constants and instead use the FetchMode constants from dbal.

we should figure out the lowest dbal version that supports those dbal constants and adjust the dbal dependency accordingly.

dbu:
fix #355 

what about `PDO::FETCH_KEY_PAIR` and `PDO::FETCH_UNIQUE | PDO::FETCH_GROUP`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,8 @@ language: php
 sudo: false
 
 php:
-  - 5.6
-  - 7.0
   - 7.1
   - 7.2
-  - hhvm
 
 env:
   - DB=mysql PACKAGE_VERSION=high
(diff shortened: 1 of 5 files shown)


Majkl578:
You will need to keep those constants from PDO for now if you rely on such behavior (but it's been deprecated).
DBAL does not support those modes natively (they are PDO-only) and support for them will likely be removed in 3.0 entirely (although KEY_PAIR mode may be supported natively in 3.0, see https://github.com/doctrine/dbal/issues/3111).

dbu:
fixed in #373

PR #361, opened 2018-08-01
making CONCAT work as expected on all platforms

rivaros:


dbu:
that looks reasonable. which database does it fix? or did you notice and wanted to do the correct thing?

the failing builds seem to be some regression from somewhere else. i restarted those builds on master. they had been green a month ago, and now the exact same thing is red :-( as all other builds are still green with this, i am ok to merge this change with those failing.

rivaros:
I guess PostgreSQL (earlier versions without CONCAT) will fail
They still use the one inherited from:
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L968

We use it with SqlServerPlatform, by forking and some XQueries
https://github.com/DataSyntax/jackalope-doctrine-dbal/commit/c07cb0ba72ea83e2c6556dd5f54ba3288ca44dc2

dbu:
thanks for the info! does the whole https://github.com/phpcr/phpcr-api-tests/ suite run with your modifications? if you want to contribute those as well, i'd be happy to review and merge them. i guess we can't run sql server tests with travis-ci? in that case i would be ok to have them without a test build, its at least a starting point for people on sql server.

dbu:
i tagged this fix at 1.3.3

PR #362, opened 2019-01-18
[BUG FIX] Fix wrong return type.

liverbool:
Force to return `array` instead of `ArrayObject` (`getNamespaces` need to return `array` by design).

https://github.com/jackalope/jackalope-doctrine-dbal/blob/058cd85e4c571f475250ad2e76d17b97feccdc75/src/Jackalope/Transport/DoctrineDBAL/Client.php#L580-L583


dbu:
i tagged 1.3.3 with this fix

PR #363, opened 2019-08-27
Use 'defaultTableOptions' to detect collate of database conection

wachterjohannes:


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -460,14 +460,17 @@ public function setCaseSensitiveEncoding($encoding)
...
+        }
 
+        $params = $this->conn->getParams();
+        $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
+        if (!isset($params['defaultTableOptions']) || !isset($params['defaultTableOptions']['collate'])) {
             return $this->caseSensitiveEncoding = $charset === 'binary' ? $charset : $charset . '_bin';
         }
 
-        return $this->caseSensitiveEncoding;
+        return $this->caseSensitiveEncoding = $params['defaultTableOptions']['collate'];
```
dbu: i'd prefer to invert this: `if (isset && isset) { return ... 'collate'; }` and then the fallback to charset binary or not.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -460,14 +460,17 @@ public function setCaseSensitiveEncoding($encoding)
      */
     private function getCaseSensitiveEncoding()
     {
-        if (!$this->caseSensitiveEncoding) {
-            $params = $this->conn->getParams();
-            $charset = isset($params['charset']) ? $params['charset'] : 'utf8';
+        if ($this->caseSensitiveEncoding) {
```
dbu: should we init this field with `null` and check `null !===` to cache the result? otherwise if we are on a non-case-sensitive encoding, we re-calculate the thing again and again for no value.

wachterjohannes:
@dbu thanks for the fast response - i have fixed all the comments

wachterjohannes:
@dbu do you have an idea why the import fails? i dont think that this has something todo with my changes

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -212,11 +212,11 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface,
     /**
      * @var string
      */
-    private $caseSensitiveEncoding;
+    private $caseSensitiveEncoding = null;
 
     /**
      * @param FactoryInterface $factory
-     * @param Connection       $conn
+     * @param Connection $conn
```
dbu: these are usually aligned. we seem to not have a codestyle thing that complains. as the parameters are declared in the method signature anyways and there is no additional information, we could also simply delete the annotations.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -373,19 +383,21 @@ public function deleteWorkspace($name)
...
         } catch (Exception $e) {
-            throw new RepositoryException("Couldn't delete workspace '$name': " .$e->getMessage(), 0, $e);
+            throw new RepositoryException("Couldn't delete workspace '$name': " . $e->getMessage(), 0, $e);
         }
 
         try {
             $this->getConnection()->delete('phpcr_nodes', ['workspace_name' => $name]);
         } catch (Exception $e) {
-            throw new RepositoryException("Couldn't delete nodes in workspace '$name': ".$e->getMessage(), 0, $e);
+            throw new RepositoryException("Couldn't delete nodes in workspace '$name': " . $e->getMessage(), 0, $e);
```
dbu: the codestyles i usually use do not put spaces around the concatenation `.`. please revert those changes to keep the diff more focussed on what this PR is doing.

i am happy to have a separate PR that does more strict codestyle (ideally with styleci configuration to keep code style strict in the future)

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -234,47 +234,54 @@ public function __construct(FactoryInterface $factory, Connection $conn)
...
+                    switch ($type) {
+                        case 'long':
+                            return (int)$content;
+                            break;
+                        case 'double':
+                            return (double)$content;
+                            break;
+                        default:
+                            return $content;
+                    }
```
dbu: while i agree with the formatting, i think such changes would better be in a separate PR that is only about styling and not mixed with feature changes, as it makes it hard to see what the actual changes are ;-)

but no need to revert it this time, as its already committed.

wachterjohannes: 👍 no problem - i have a built in formatter in my ide so its hard to seperate it! but will take a look at it next time

wachterjohannes:
@dbu thanks for merging it - when can i expect a release to happen?

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: yeah i thought as much. for reviewing its not ideal, and neither for git history to figure out what was going on ;-)

dbu:
i just tagged 1.3.4 with this (and no other changes) :checkered_flag:

wachterjohannes:
🥇 thanks -

PR #364, opened 2019-10-23
fix nested joins where clause generation

esserj:
when having nested joins the current QOMWalking is falsly
assuming the right join as the most left join resulting in
a wrong query when having multiple joins with different
node_types. included tests highlight the problem.

i have a document and another document that's a weak reference,
I then join the weakReference but also need to join a child document
if my document now has a type that the child doesn't have the resulting
query will falsely assume the root type is that of the weakReference joined because
its using the right alias alias of the left join instead of the most left document (left alias of the left join).

is generates:
```sql
WHERE n1.workspace_name = 'default'
AND n1.type IN ('nt:unstructured', 'rep:root','mycustomtype')
```

while it should be:
```sql
WHERE n0.workspace_name = 'default'
AND n0.type IN ('nt:unstructured', 'rep:root','mycustomtype')
```

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -315,6 +315,41 @@ class QOMWalker
         throw new BadMethodCallException('Supplied join type should implement getSelector2Name() or be an instance of ChildNodeJoinConditionInterface or DescendantNodeJoinConditionInterface');
     }
 
+
+    /**
+     * @param QOM\JoinConditionInterface $right
+     *
+     * @return string the alias on the left side of a join
+     *
+     * @throws BadMethodCallException if the provided JoinCondition has no valid way of getting the left selector
+     */
+    private function getLeftJoinSelector(QOM\JoinConditionInterface $left)
+    {
+        if ($left instanceof QOM\ChildNodeJoinConditionInterface) {
+            return $left->getChildSelectorName();
+        } elseif ($left instanceof QOM\DescendantNodeJoinConditionInterface) {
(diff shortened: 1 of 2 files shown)


esserj:
tests failing are unrelated

dbu:
thanks a lot for digging through the code to find this issue! i will tag a release shortly.

a word of caution: phpcr is not well suited to do complex queries. if you need complex queries and have a straightforward data structure, a direct relational approach will be more sustainable.

the main use case of phpcr is to use the document tree directly. it brings value when you have nested trees of data with varying types of documents in the tree.

esserj:
Hi @dbu I agree, but we build a CMS based on symfony CMF the queries we do in this case was our backend UI search that was searching through translated text for matches of a user search query, and the tree displaying the data already had categories joined for display purposes, that's how this came to light.

in the front we do filtering on publication information. Not sure how one could build a full functional CMS in PHPCR without any queries really or offload the work to php but either way have a performance hit when no cache is available.

esserj:
anyways, thanks alot for the quick response and merge!

dbu:
@esserj sorry, i did not want to say "you are doing it wrong". just noticed that sometimes people try to build a relational data structure on top of phpcr, which i would generally consider a bad idea. it is totally fine to have some need for queries (and thats what the query system is there for, after all :-) ) and when you do queries, you expect them to be correct, so thanks for the fix!

esserj:
@dbu wasn't taking it as "ur doing it wrong" :) was actually hoping to be enlightened with a better way of doing certain things in context of a cms :D always open to learn from ppl with more experience than I, it is our first time in the world of PHPCR so there is alot to be learned

PR #367, opened 2019-12-10
Fix reordering of namespaced children fails

esserj:
We are running into an issue where reordering namespaced nodes fails because of the following issue:

When you have 2 namespaced nodes under the same parent `Client::reorderChildren` is trying to sort them by concatenating namespace and nodename, the namespace in de nodes table is stored as the uri of the namespace instead of the prefix (namespace table primary key). In several locations in the client the uri is fetched and used for nodes queries, but in the reorder is it not.

The related api test in OrderBeforeTest::testNodeOrderNamespaces passed because the `DBUnitFixturesXML` was importing the prefix into the namespace instead of the uri (as the library is doing).

Fixing both Client and import now passes these tests and ensures namespaced nodes get reordered as expected

### DIFF
--- a/tests/Jackalope/Test/Fixture/DBUnitFixtureXML.php
+++ b/tests/Jackalope/Test/Fixture/DBUnitFixtureXML.php
@@ -160,8 +160,8 @@ class DBUnitFixtureXML extends XMLDocument
 
         $dom = new DOMDocument('1.0', 'UTF-8');
         $phpcrNode = $dom->createElement('sv:node');
-        foreach ($this->namespaces as $namespace => $uri) {
-            $phpcrNode->setAttribute('xmlns:' . $namespace, $uri);
+        foreach ($this->namespaces as $namespacePrefix => $uri) {
+            $phpcrNode->setAttribute('xmlns:' . $namespacePrefix, $uri);
         }
         $dom->appendChild($phpcrNode);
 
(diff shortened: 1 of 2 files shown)


tests/Jackalope/Test/Fixture/DBUnitFixtureXML.php
```diff
@@ -192,6 +192,9 @@ public function addNode($workspaceName, DOMElement $node)
             $name       = '';
             $namespace  = '';
         }
+        if (isset($this->namespaces[$namespace])) {
```
dbu: for better readability, could we rename the above $namespace to $namespacePrefix and then change this to `$namespace = isset($this->namespaces[$namespace]) ? $this->namespaces[$namespacePrefix] : $namespacePrefix;` (the `??` operator is not available in all PHP versions we currently support, unfortunately).

esserj: sure, i'll get that in right away.

The tests I'll see when and if I can make some time for it, but there is another PR i'm gona make for a similar bug as the incorrect nested joins where clause

esserj: @dbu not sure if you're code suggestion is ok (apart from the type in the isset $namespace instead of $namespacePrefix):
`$namespace = isset($this->namespaces[$namespace]) ? $this->namespaces[$namespacePrefix] : $namespacePrefix;`

I don't think you can fallback here on $namespacePrefix, technically you should even throw an exception if $namespacePrefix is set and not found in namespaces.

dbu: ups, yeah i mixed up things. but you are right - as this is in tests we don't have to worry about BC breaks, lets do

```
if (!isset($this->namespaces[$namespacePrefix])) {
    throw new \Exception('Undefined namespace prefix '.$namespacePrefix);
}
and then directly use `$this->namespaces[$namespacePrefix]` in the code below?
```

dbu: ah i see you pushed. forgot about the special case... i like what you have now, seems good to me.

do things in the jcr namespace not have a namespace url?

esserj: I'm not sure if they do, not to familiar with the spec, I do know that in the tests they use jcr:one and that is expected to be namespaced, but the root node seems exempt? Not sure why.

dbu: and i don't remember :-) probably just historic reasons from the beginning. i don't want to mess with root though, so its fine now - thanks again!

PR #369, opened 2019-12-11
cleanup build

dbu:


dbu:
sqlite seems to have a pretty consistent issue with that foreign key error during import.

php 7.4 can not build because we would need to run with a newer phpunit.

dbu:
@esserj i fix most of the build issues in this change \o/

what is left is something with sqlite foreign key handling instability (that fails on and off and seems to not be a problem for mysql nor postgres) and we would need to go to newer phpunit versions to test with php 7.4. phpunit 5 || 6 should hopefully cover all versions without too many bc breaks.

.travis.yml
```diff
@@ -1,43 +1,38 @@
...
 
 services:
   - mysql
   - postgresql
 
 cache:
   directories:
-    - vendor
-    - $HOME/.composer/cache
+          #    - $HOME/.composer/cache
```
dbu: revert when build succeeds

composer.json
```diff
@@ -26,9 +27,9 @@
     },
     "require-dev": {
         "psr/log": "^1.0",
-        "phpcr/phpcr-api-tests": "2.1.19",
-        "phpunit/phpunit": "^5.7 || ^6.4",
-        "phpunit/dbunit": "^1.3"
+        "phpcr/phpcr-api-tests": "dev-phpunit-version@dev",
```
dbu: tag and cleanup when ready.

PR #371, opened 2019-12-20
Fix move with namespaced nodes

esserj:
Hi @dbu me **again** :)

When you currently move a namespace'd node, the query build will change the local_name to the local_name including namespace while everywhere else it is stored without the namespace.

I've added a test first to prove current implementation is broken and the follow-up commit shows its now working as expected.

Thanks again for making time to review this!


dbu:
looks like nobody had been moving nodes outside of default namespace with jackalope-doctrine-dbal... thanks for the fix!

this smells like we also miss a test for this case in https://github.com/phpcr/phpcr-api-tests/blob/master/tests/Writing/MoveMethodsTest.php

would be cool if you want to contribute that test too. i think the one in here should still be kept as we can more specifically check with a dbal query if internal state is fine.

dbu:
tagged 1.4.1 with this fix

esserj:
Hi @dbu  i'm gona put that on my todo list because it does indeed belong there.
Thanks for the quick merge and release!

PR #372, opened 2020-08-04
feat(primary key): add missing primary key for phpcr_type_childs

MichelFalke:
If settings for percona or strict it requires a primary key on each table. To be backwards compatible we added a primary key on a new field, so there won't be breaking changes.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,14 @@
 Upgrade
 =======
 
+1.5
+---
+
+- The following change needs to be made to your database: 
+
+  `ALTER TABLE phpcr_type_childs ADD COLUMN `id` INT(11) UNSIGNED PRIMARY KEY AUTO_INCREMENT FIRST;` 
+  
+
 1.2
 ---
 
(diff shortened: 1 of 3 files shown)


UPGRADE.md
```diff
@@ -1,6 +1,14 @@
 Upgrade
 =======
 
+1.5
+---
+
+- The following change needs to be made to your database:
```
dbu, line 7:
```suggestion
- If you want to start using Percona in strict mode, or use the Percona cluster, add a primary key to the phpcr_type_childs:
```

MichelFalke:
I think it will add functionality for the Percona cluster and will not break any other situation. Thats why my opinion on this matter is that it shouldn't be more complex than this PR. Thanks for the feedback on the upgrade.

alexander-schranz:
We are also running in digitalocean into this issue. Is it possible to create a release for this. Think it will break nothing as we don't do any queries on the id.

dbu:
sure: https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.5.0

PR #375, opened 2020-10-18
prevent type error on Postgresql

RobertMe:
Prevent a cast error on Postgresql when no target is found for the given path. This by adding an early return of an empty array. MySQL and possibly other databases would cast the false value to 0, after which the query result would still be an enpty array as no node can have a target_id equaling 0.


RobertMe:
I've run into this error when using Sulu CMS running on a Postgresql database. I think this is the correct fix, although I didn't dig into it too much. At least I found it odd that the query is run when there is no targetId, so I believe this early return with an empty array would be correct.

dbu (approved):
thanks, i think not querying in that case makes a lot of sense - also performance wise.

dbu:
https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.5.1

PR #377, opened 2020-12-04
Add support for dbal 3 (and php 8)

alekitto:
I've finished the work started in `php8` branch adding support for dbal 3.

DBUnit has been removed from dependencies, and an `XMLDataSet` class simulating its functionalities has been added to the `Jackalope\Test\Tester` namespace.
To correctly handle BC breaks introduced in DBAL, the minimum version of `doctrine/dbal` has been raised to 2.11. Consequently the minimum PHP version now is 7.3.

Missing fetch mode are simulated via `array_*` functions to maintain the expected logic.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,23 +15,23 @@
         }
     ],
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^7.3 || ^8.0",
         "ext-pdo": "*",
-        "doctrine/dbal": "^2.9.3",
+        "doctrine/dbal": "^2.11 || ^3.0",
         "phpcr/phpcr": "~2.1.5",
-        "phpcr/phpcr-utils": "^1.4.1",
-        "jackalope/jackalope": "dev-phpunit-9@dev"
+        "phpcr/phpcr-utils": "^1.5.0",
+        "jackalope/jackalope": "^1.4.2"
     },
     "provide": {
         "jackalope/jackalope-transport": "1.3.0"
     },
     "require-dev": {
         "psr/log": "^1.0",
-        "phpcr/phpcr-api-tests": "dev-allow-php8@dev",
-        "phpunit/phpunit": "^7.5.18 || ^8.0 || ^9.0",
-        "phpunit/dbunit": "^4.0"
+        "phpcr/phpcr-api-tests": "2.1.22",
+        "phpunit/phpunit": "^9.0"
     },
     "autoload": {
+        "files": [ "src/dbal2_compat.php" ],
         "psr-0": { "Jackalope\\": "src/" }
     },
     "autoload-dev": {
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -7,8 +7,12 @@ use Closure;
 use DateTime;
 use DateTimeZone;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Driver\PDO\Connection as PDOConnection;
+use Doctrine\DBAL\Exception as DBALException;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use DOMDocument;
 use DOMElement;
@@ -34,7 +38,6 @@ use Jackalope\Transport\TransactionInterface;
 use Jackalope\Transport\WorkspaceManagementInterface;
 use Jackalope\Transport\WritingInterface;
 use PDO;
-use PDOException;
 use PHPCR\AccessDeniedException;
 use PHPCR\CredentialsInterface;
 use PHPCR\ItemExistsException;
@@ -227,12 +230,8 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
     /**
      * @TODO: move to "SqlitePlatform" and rename to "registerExtraFunctions"?
-     *
-     * @param PDO $sqliteConnection
-     *
-     * @return Client
      */
-    private function registerSqliteFunctions(PDO $sqliteConnection)
+    private function registerSqliteFunctions(PDO $sqliteConnection): void
     {
         $sqliteConnection->sqliteCreateFunction(
             'EXTRACTVALUE',
@@ -261,10 +260,10 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
                     switch ($type) {
                         case 'long':
                             return (int)$content;
-                            break;
+
                         case 'double':
                             return (double)$content;
-                            break;
+
                         default:
                             return $content;
                     }
@@ -282,8 +281,6 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
                 return implode('', func_get_args());
             }
         );
-
-        return $this;
     }
 
     /**
@@ -491,9 +488,9 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     {
         try {
             $query = 'SELECT 1 FROM phpcr_workspaces WHERE name = ?';
-            $result = $this->getConnection()->fetchColumn($query, [$workspaceName]);
+            $result = $this->getConnection()->fetchFirstColumn($query, [$workspaceName]);
         } catch (Exception $e) {
-            if ($e instanceof DBALException || $e instanceof PDOException) {
+            if ($e instanceof DBALException) {
                 if (1045 === $e->getCode()) {
                     throw new LoginException('Access denied with your credentials: ' . $e->getMessage());
                 }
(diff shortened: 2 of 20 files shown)


tests/ImplementationLoader.php
```diff
@@ -192,15 +195,21 @@ public function getRepository()
 
     public function getFixtureLoader()
     {
-        $testerClass = '\\Jackalope\\Test\\Tester\\' . ucfirst(strtolower($this->connection->getWrappedConnection()->getAttribute(PDO::ATTR_DRIVER_NAME)));
-        if (!class_exists($testerClass)) {
-            // load Generic Tester if no database specific Tester class found
-            $testerClass = Generic::class;
+        $platform = $this->connection->getDatabasePlatform();
+        switch (true) {
```
dbu, line 199:
```suggestion
        switch ($platform) {
```

semantically the same, but i find it nice to specify the thing that we switch on here

tests/Jackalope/Test/Tester/Generic.php
```diff
@@ -51,32 +52,48 @@ public function getConnection()
...
+        $platform = $this->connection->getDatabasePlatform();
+        foreach ($this->dataSet->getTables() as $table) {
+            $this->connection->executeStatement($platform->getTruncateTableSQL($table->getName(), true));
+        }
+
+        foreach ($this->dataSet->getTables() as $table) {
+            foreach ($this->dataSet->getRows($table->getName()) as $row) {
+                $sql = 'INSERT INTO '.$platform->quoteIdentifier($table->getName()).
+                    ' ('.\implode(',', array_keys($row)).') VALUES ('.
+                    implode(',', array_fill(0, count($row), '?')).')';
```
dbu, line 93:
```suggestion
                    \implode(',', array_fill(0, count($row), '?')).')';
```

tests/Jackalope/Tools/Console/InitDoctrineDbalCommandTest.php
```diff
@@ -45,7 +45,7 @@ public function setUp(): void
         $this->connection = $this->createMock(Connection::class);
         $this->schemaManager = $this->createMock(AbstractSchemaManager::class);
 
-        $this->platform = $this->createMock(MySqlPlatform::class);
+        $this->platform = $this->createMock(MySQLPlatform::class);
```
dbu, line 48: the case change seems to happen later, there is a warning in the lowest version build: https://travis-ci.com/github/jackalope/jackalope-doctrine-dbal/jobs/454621052

dbal 2 seems to still have it lowercase. would it make sense to define an alias if the uppercase version does not exist?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -7,8 +7,11 @@
 use DateTime;
 use DateTimeZone;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\DBALException;
-use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\DBAL\Driver\PDO\Connection as PDOConnection;
+use Doctrine\DBAL\Exception as DBALException;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
```
dbu, line 13: it seems that dbal 2 provides only the MySqlPlatform with lower case. i think we should define an alias if MySQLPlatform does not exist to avoid regressions with dbal 2.

tests/ImplementationLoader.php
```diff
@@ -1,18 +1,21 @@
 <?php
 
+use Doctrine\DBAL\Platforms\MySQLPlatform;
```
dbu, line 3: same here, alias for dbal 2?

dbu (commented):
wow, thanks a lot for that work, great job!

styleci has a tiny complaint, can you please fix that extra line?

and .phpunit.result.cache looks like it should not be committed. can you please remove that again and add it to .gitignore?

i added a few minor comments:

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -227,10 +229,6 @@ public function __construct(FactoryInterface $factory, Connection $conn)
 
     /**
      * @TODO: move to "SqlitePlatform" and rename to "registerExtraFunctions"?
-     *
-     * @param PDO $sqliteConnection
-     *
-     * @return Client
```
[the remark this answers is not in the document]
alekitto: I there any value to return `$this` at the end of this method? `: void` as return type could be added otherwise?

composer.json
```diff
@@ -15,21 +15,20 @@
...
     "provide": {
         "jackalope/jackalope-transport": "1.3.0"
     },
     "require-dev": {
         "psr/log": "^1.0",
-        "phpcr/phpcr-api-tests": "dev-allow-php8@dev",
-        "phpunit/phpunit": "^7.5.18 || ^8.0 || ^9.0",
-        "phpunit/dbunit": "^4.0"
+        "phpcr/phpcr-api-tests": "^2.1.22",
+        "phpunit/phpunit": "^7.5.18 || ^8.0 || ^9.0"
```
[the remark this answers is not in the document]
alekitto: I've just discovered that phpunit 7 is necessary to correctly execute tests with lower dependencies: `Jackalope\TestCase::setUp` method was missing the return type in a previous version.

src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
```diff
@@ -922,7 +922,7 @@ private function sqlXpathValueExists($alias, $property)
             return "EXTRACTVALUE($alias.props, 'count(//sv:property[@sv:name=\"" . $property . "\"]/sv:value[1])') = 1";
         }
 
-        if ($this->platform instanceof PostgreSqlPlatform) {
+        if ($this->platform instanceof PostgreSQL94Platform) {
```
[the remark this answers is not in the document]
alekitto: You're right. I've re-added the checks on `PostgreSqlPlatform`

tests/Jackalope/Tools/Console/InitDoctrineDbalCommandTest.php
alekitto: I was really short of good ideas here. Unfortunately `MySqlPlatform` and `MySQLPlatform` are exactly the same token for PHP, as class names are case-insensitive. The problem here is `require` the correct file on case-sensitive filesystems.

I've added a `dbal2_compat.php` file which adds a simple autoloader function in case the uppercase-named class was not present, but I'm unsure about that. What do you think about it?

tests/Jackalope/Test/Tester/Generic.php
alekitto: I've imported the function with `use function` instead.

composer.json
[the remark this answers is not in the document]
dbu: can we go for a newer version of jackalope/jackalope to avoid that? imho there is no need to be very open with the jackalope/jackalope versions we allow.

composer.json
```diff
@@ -15,23 +15,23 @@
...
     },
     "provide": {
         "jackalope/jackalope-transport": "1.3.0"
     },
     "require-dev": {
         "psr/log": "^1.0",
-        "phpcr/phpcr-api-tests": "dev-allow-php8@dev",
-        "phpunit/phpunit": "^7.5.18 || ^8.0 || ^9.0",
-        "phpunit/dbunit": "^4.0"
+        "phpcr/phpcr-api-tests": "^2.1.22",
```
dbu, line 30:
```suggestion
        "phpcr/phpcr-api-tests": "2.1.22",
```

the tests are not really semver - if we add a new test, it might fail for dbal. we lock to a specific tag of the test suite.

src/Jackalope/Transport/DoctrineDBAL/Client.php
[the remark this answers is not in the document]
dbu: oh, you are right. that is better :+1: 
thanky ou

tests/Jackalope/Tools/Console/InitDoctrineDbalCommandTest.php
dbu: clever solution! i asked on twitter if there is any best practice but at least this would work.

dbu: on twitter, somebody said it did [not work for him](https://twitter.com/rpkamp82/status/1335573797383139329), guessing that class_exists within an autoloader does not attempt an autoload to prevent endless loops.

makes me wonder if it would be enough to have something (outside the autoloaders) that does `class_exists(MySQLPlatform::class) || class_exists(MySqlPlatform::class)`? that dbal2_compat.php could maybe have just those 2 lines, no? if it is always executed, it will load the class if it exists at all and then case sensitivity of the filesystem no longer matters.

alekitto: Mh, that's weird, probably WSL is involved in this error. Anyway I updated the code to call two class_exists in the compat file as you suggested.

composer.json
alekitto: 👍

src/dbal2_compat.php
```diff
@@ -0,0 +1,3 @@
+<?php declare(strict_types=1);
+
+class_exists('Doctrine\DBAL\Platforms\MySQLPlatform') || class_exists('Doctrine\DBAL\Platforms\MySqlPlatform');
```
dbu, line 3:
```suggestion
// doctrine/dbal version 2 used the lowercase form. On case sensitive filesystems, this leads to autoloading not working
// for PHP, class names are case insensitive.
class_exists('Doctrine\DBAL\Platforms\MySQLPlatform') || class_exists('Doctrine\DBAL\Platforms\MySqlPlatform');
```

dbu:
thanks a lot! i will wrap up the changelog and tag a release :+1:

PR #379, opened 2021-04-12
[WIP] Add: phpcr-utils PR test

flogado:
Temporary pull request for test purposes.

See: https://github.com/phpcr/phpcr-utils/pull/194

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "ext-pdo": "*",
         "doctrine/dbal": "^2.11 || ^3.0",
         "phpcr/phpcr": "~2.1.5",
-        "phpcr/phpcr-utils": "^1.5.0",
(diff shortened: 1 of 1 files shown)


dbu:
thanks. this looks pretty good, except for the lowest version build. i am quite confused by that one as we have other tests in phpcr-utils that also extend from the TestCase and those don't fail. i just rebuilt the master branch and it is green too. i will go ahead and merge - worst case we need to dig into what goes wrong with this specific test in the lowest version build.

dbu:
i merged and tagged the phpcr-utils change. lets see what the lowest version build does now.

dbu:
not sure what exactly happened, but i re-ran the build on the master branch after tagging 1.5.2 and all builds succeed \o/

thanks again for the fix.

PR #380, opened 2021-04-13
CachedClient: Intepret only FALSE as a cache miss

playr1983g:
Hi. while exploring the possibility of using query cache in our project, we noticed that empty results were not interpreted as a cache hit. We added a stricter check (!== false) to make this possible.

This point is somewhat related to this issue: https://github.com/doctrine/DoctrinePHPCRBundle/pull/359


dbu (approved):
thanks!

dbu:
released as 1.6.1

playr1983g:
thanks @dbu 
incredibly fast :)

PR #381, opened 2021-05-31
Improve performance for xml parsing

alexander-schranz:
I currently did analyse the performance of @sulu in https://github.com/sulu/sulu/pull/6089 after some optimizations on our side most time is spend now in `xmlToColumns` see the following benchmark:

## Before

Response time ~1080ms

 - https://blackfire.io/profiles/af73cb80-e327-4d12-9e76-d729b7097b86/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()
 
#### Apache Bench Test:
 
 ```bash
ab -n 50 -c 1 -l http://localhost/my-uri
 ```
 
 <details>
 <summary>Result: Requests per second:    1.01 [#/sec] (mean)</summary>
 
 ```bash
 Server Software:        nginx/1.21.0
Server Hostname:        stern-nl.localhost
Server Port:            80

Document Path:          /my-uri
Document Length:        348158 bytes

Concurrency Level:      1
Time taken for tests:   67.569 seconds
Complete requests:      50
Failed requests:        0
Total transferred:      17429509 bytes
HTML transferred:       17407759 bytes
Requests per second:    0.74 [#/sec] (mean)
Time per request:       1351.384 [ms] (mean)
Time per request:       1351.384 [ms] (mean, across all concurrent requests)
Transfer rate:          251.90 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   974 1351 479.1   1156    3550
Waiting:      973 1344 474.9   1155    3548
Total:        974 1351 479.1   1156    3550

Percentage of the requests served within a certain time (ms)
  50%   1156
  66%   1399
  75%   1533
  80%   1590
  90%   1896
  95%   2427
  98%   3550
  99%   3550
 100%   3550 (longest request)
 ```
 
 </details>
 
Bildschirmfoto 2021-06-01 um 01 36 17

 
## After

Response time ~580ms
 
#### Apache Bench Test:
 
 ```bash
ab -n 50 -c 1 -l http://localhost/my-uri
 ```
 
 <details>
 <summary>Result: Requests per second:    1.91 [#/sec] (mean)</summary>
 
 ```bash
Server Software:        nginx/1.21.0
Server Hostname:        stern-nl.localhost
Server Port:            80

Document Path:          /my-uri
Document Length:        348161 bytes

Concurrency Level:      1
Time taken for tests:   30.322 seconds
Complete requests:      50
Failed requests:        0
Total transferred:      17429472 bytes
HTML transferred:       17407722 bytes
Requests per second:    1.65 [#/sec] (mean)
Time per request:       606.450 [ms] (mean)
Time per request:       606.450 [ms] (mean, across all concurrent requests)
Transfer rate:          561.33 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   504  606  88.9    571     789
Waiting:      503  605  88.8    570     789
Total:        504  606  88.9    571     789

Percentage of the requests served within a certain time (ms)
  50%    571
  66%    676
  75%    693
  80%    700
  90%    729
  95%    774
  98%    789
  99%    789
 100%    789 (longest request)
 ```
 
 </details>

  - https://blackfire.io/profiles/30ffb1b2-540d-4a7b-8831-68d3fe6c4564/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()
 
Bildschirmfoto 2021-06-01 um 01 35 43
 
----
 
 This PR tries to use an more efficient way of parsing the xml structure sadly in costs of readability of the code, hope I can get into a form which it would be still maintainable.
 
 The PR is not yet finished and I'm just experimenting here a little bit. The blackfire profiles seems currently not show the result I'm seeing in the response times, need to investigate here more time. Update added the apache bench benchmark results for better performance testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.7.0
+-----
+
+* Improve performance for xml parsing.
+
 1.6.1
 -----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Changelog
 1.7.0
 -----
 
-* Improve performance for xml parsing.
+* Improve performance for xml parsing by using the PHP ext-xml instead of ext-dom.
+  Jackalope\Transport\DoctrineDBALClient::mapPropertyFromElement is no longer called within the client. If you extended the client and call the method, things will still work as before, but it is recommended to refactor your code to use the XmlToPropsParser. If you overwrote the method behaviour, your changes will no longer happen because the method is not called anymore.
 
 1.6.1
 -----
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -30,6 +30,7 @@ use Jackalope\Query\QOM\QueryObjectModelFactory;
 use Jackalope\Query\Query;
 use Jackalope\Transport\BaseTransport;
 use Jackalope\Transport\DoctrineDBAL\Query\QOMWalker;
+use Jackalope\Transport\DoctrineDBAL\XmlParser\XmlToPropsParser;
 use Jackalope\Transport\MoveNodeOperation;
 use Jackalope\Transport\NodeTypeManagementInterface;
 use Jackalope\Transport\QueryInterface as QueryTransport;
@@ -1096,16 +1097,12 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
      */
     protected function xmlToProps($xml)
     {
-        $data = new stdClass();
+        $xmlParser = new XmlToPropsParser(
+            $xml,
+            $this->valueConverter
+        );
 
-        $dom = new DOMDocument('1.0', 'UTF-8');
-        $dom->loadXML($xml);
-
-        foreach ($dom->getElementsByTagNameNS('http://www.jcp.org/jcr/sv/1.0', 'property') as $propertyNode) {
-            $this->mapPropertyFromElement($propertyNode, $data);
-        }
-
-        return $data;
+        return $xmlParser->parse();
     }
 
     /**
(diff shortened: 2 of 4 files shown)


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1124,20 +1124,97 @@ protected function xmlToProps($xml)
...
-        $dom = new DOMDocument('1.0');
-        $dom->loadXML($xml);
-        $dom->formatOutput = true;
-        $xpath = new DOMXPath($dom);
 
-        foreach ($propertyNames as $propertyName) {
-            $els = $xpath->query(sprintf('.//sv:property[@sv:name="%s"]', $propertyName));
-            if ($els->length === 0) {
-                continue;
+        $lastPropertyName = null;
```
alexander-schranz, line 1128: this could be moved into a class then the callback method can access this properties with `$this->lastPropertyName` instead of using `&` must look into it what performance impact that will have.

alexander-schranz:
@dbu what do you think in general about changing the xml parsing here?

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: i like that idea, also because the Client is already huge and adding this inside a method does not improve readability ;-)

dbu (approved):
very cool that you found such a performance improvement!

i was wondering if we need to change our composer require because of this, but it seems that it was wrong because ext-dom is not required. ext-xml is required in jackalope/jackalope, so i think we already are good on that one.

would be cool if you can extract the code into a class that the client can instantiate to not further increase the size of the client. maybe you can even add some tests directly with that new parser class?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1124,20 +1124,97 @@ protected function xmlToProps($xml)
...
+        $currentValues = [];
+        $currentValue = null;
+
+        $parser = xml_parser_create();
+        xml_set_element_handler(
+            $parser,
+            function ($parser, $name, $attrs) use ($propertyNames, &$lastPropertyName, &$lastPropertyType, &$lastPropertyMultiValued, &$currentTag) {
+                $currentTag = $name;
+
+                if ($name === 'SV:PROPERTY') {
```
dbu, line 1141: in the interest of early return, we could say `if ('SV:PROPERTY' !== $name) { return`

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1124,20 +1124,97 @@ protected function xmlToProps($xml)
...
+                            $data->{$lastPropertyName} = $lastPropertyMultiValued ? $currentValues : $currentValues[0];
+                            $data->{':' . $lastPropertyName} = $lastPropertyType;
+                            break;
+                    }
+
+                    $currentValues = [];
+                    $lastPropertyName = null;
+                    $lastPropertyType = null;
+                    $lastPropertyMultiValued = null;
+                } elseif ($name === 'SV:VALUE' && $lastPropertyName) {
```
dbu, line 1166: can we return abvoe, instead of the elseif?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1124,20 +1124,97 @@ protected function xmlToProps($xml)
...
+                            break;
+                        case PropertyType::DOUBLE:
+                            $currentValues[] = (double)$currentValue;
+                            break;
+                        default:
+                            throw new InvalidArgumentException("Type with constant $type not found.");
+                    }
+                }
+
+                $currentTag = null;
```
dbu, line 1204: if i understand correctly, this could also be at the top of the function, it is to reset the value set by the other function, right? if we do early return we would need to move it up.

alexander-schranz: yes as we are not accessing it inside this method we can set `$currentTag = null;` at beginning

src/Jackalope/Transport/DoctrineDBAL/Client.php
alexander-schranz: sure will do this :)

alexander-schranz:
@dbu Thank you for the review will try to do the changes at the weekend!

alexander-schranz:
@dbu Did now create a class and use it inside `xmlToProps` and `xmlToColumns` methods. Found some edge cases I didn't take into account (sv:property without sv:value, concated data values when having encoded string in the xml (foo &amp; bar). Basically they did fail in the whole test suit, still I created additional unit test for the new class as it was easier for testing. Pull Request is from my side now ready for review.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1146,6 +1133,8 @@ protected function xmlToColumns($xml, $propertyNames)
      * @param stdClass $data
      *
      * @throws InvalidArgumentException
+     *
+     * @deprecated This method was replaced with the `XmlToPropsParser` for performance reasons.
```
alexander-schranz, line 1137: This method is actually not longer called and if somebody did override it will have no effect, should we remove it. Would be a little bc break but actually don't think somebody used this method.

src/Jackalope/Transport/DoctrineDBAL/XmlParser/XmlToPropsParser.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * @var string[]|null
+     */
+    private $propertyNames;
+
+    /**
+     * @var ValueConverter
+     */
+    protected $valueConverter;
```
dbu, line 26: what is the use case of having this property be `protected` instead of `private`? i suggest to make this private as well and mark the whole class as `final`.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: hm. if somebody extended the client and called this method, then it would be ok to keep it as you did now.

if somebody extended the method, their changed method will no longer be called. that is already a BC break. though i agree with you it is unlikely anyone is using this.

i suggest keeping it as you now did it, with the deprecation annotation.

dbu (approved):
great job, thanks a lot!

can you please add a line to the CHANGELOG.md, for a next minor version?

src/Jackalope/Transport/DoctrineDBAL/XmlParser/XmlToPropsParser.php
alexander-schranz: copy paste error 😄 will change it.

src/Jackalope/Transport/DoctrineDBAL/Client.php
alexander-schranz: I did also add a `@trigger` so it would also be logged.

alexander-schranz:
@dbu Code updated :)

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+1.7.0
+-----
+
+* Improve performance for xml parsing.
```
dbu, line 7:
```suggestion
* Improve performance for xml parsing by using the PHP ext-xml instead of ext-dom.
  Jackalope\Transport\DoctrineDBALClient::mapPropertyFromElement is no longer called within the client. If you extended the client and call the method, things will still work as before, but it is recommended to refactor your code to use the XmlToPropsParser. If you overwrote the method behaviour, your changes will no longer happen because the method is not called anymore.
```

dbu (approved):
thanks a lot!

dbu:
https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.7.0

alexander-schranz:
@dbu Thank you!

plozmun:
Hi @alexander-schranz , i've tested this change and it is a great improvement. Thank you!!! 👏

alexander-schranz:
@plozmun thank you for testing this out. I'm curious how much the performance difference it is in your case for uncached pages in prod env.

plozmun:
> @plozmun thank you for testing this out. I'm curious how much the performance difference it is in your case for uncached pages in prod env.

Frontend performance has improved over 70-80% for uncached pages

In the admin is where the improvement is huge, from having some timeout to responses in few seconds for sections with many pages.

alexander-schranz:
@plozmun Wow 70-80% this is really great, would not have expected that 😲. So your timeouts issues are now gone? Which sulu version did you use to test it?

plozmun:
@alexander-schranz yes! We have pages with many relations and 20 languages and those pages are where it really improves. Sulu version is 2.3

alexander-schranz:
@plozmun Really great to hear that this did also solve your performance issues. Think with this stats you did use the `2.3@dev` version wich also includes https://github.com/sulu/sulu/pull/6089 as this also improve the performance when loading nodes.
Thank you for your feedback and testing this so out so early 🙏.

PR #385, opened 2021-09-21
make code more robust

dbu:


danrot:
@dbu Not working for Sulu anymore 😬 I guess you should ask @nnatter or @alexander-schranz

composer.json
```diff
@@ -15,10 +15,10 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.4 || ^8.0",
         "ext-pdo": "*",
         "ext-dom": "*",
-        "doctrine/dbal": "^2.11 || ^3.0",
+        "doctrine/dbal": "^3.1",
```
alexander-schranz, line 21:
```suggestion
        "doctrine/dbal": "^2.13 || ^3.1",
```

Would be good to support also dbal 2 as doctrine/orm still requires dbal 2

dbu:
sorry for the noise, dan.

> I think there will be no issue if we will allow both versions ^1.3 || ^2.0 in our composer.json. 

indeed, there are no changes to PHPCR planned, so for consumers there should be no BC breaks. i plan to introduce static typing where possible, and the changes are large enough that i think we should bump the major version because there could be different behaviour in edge cases.

> The problem is currently more the upgrade to "doctrine/dbal": "^3.1", as doctrine/orm still requires 2.0 so I can not yet test sulu with this branch.

uh, good catch, that would not be good. i will relax the version constraints again. i got tricked into thinking it would be quite different, but actually the only difference seems renaming Statement::execute to Statement::executeQuery, which we use in one place...

alexander-schranz:
@dbu I got the following error on linting the stage environment container:

```
Cache provided for type "meta" is Symfony\Component\Cache\DoctrineProvider   
which does not implement Psr\SimpleCache\CacheInterface.                     
```

dbu:
ah right, that is to be expected. that was the one BC break i wanted to do (because doctrine cache has been abandoned).

do you use https://github.com/doctrine/DoctrinePHPCRBundle to boostrap jackalope, or your own code? i think this will be a thing where bootstrapping needs to check the jackalope version to make this work. i notice that we do not have a version constant in jackalope-doctrine-dbal. we do have one in jackalope-jackrabbit on the `Client`. would it make sense to add a version constant in the jackalope-doctrine-dbal Client as well? you'd have to check with [defined](https://www.php.net/manual/en/function.defined.php) if the constant is there at all, as we don't have it in version 2.

PR #386, opened 2021-09-23
upgrade to jackalope 2

dbu:


tests/ImplementationLoader.php
```diff
@@ -203,7 +203,7 @@ public function getFixtureLoader()
                 $testerClass = Mysql::class;
                 break;
 
-            case $platform instanceof PostgreSQL94Platform || $platform instanceof PostgreSqlPlatform:
+            case $platform instanceof PostgreSQL94Platform || $platform instanceof PostgreSQLPlatform:
```
alexander-schranz, line 206: alternative I think we could check for `$platform->getName()` instance on building on FQCN. 

https://github.com/doctrine/dbal/blob/96b0053775a544b4a6ab47654dac0621be8b4cf8/src/Platforms/PostgreSQL94Platform.php#L1067

https://github.com/doctrine/dbal/blob/96b0053775a544b4a6ab47654dac0621be8b4cf8/src/Platforms/MySQLPlatform.php#L1039

PR #389, opened 2021-10-21
PHP 7.2 compatibility added

d3vpunk:
composer.json:
- changed minimal PHP version to 7.2
- Fixed to 1.x version of doctrine/cache
- Changed PHPUnit version from 9.0 back to 8.5 (PHP 7.2 compatible)
Test matrix:
- Added PHP 7.2 to test matrix

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -33,15 +33,6 @@ jobs:
                 dependencies:
                     - highest
                 include:
-                    - php-version: '7.3'
-                      dependencies: lowest
-                      db: sqlite
-                    - php-version: '7.3'
-                      dependencies: lowest
-                      db: mysql
-                    - php-version: '7.3'
-                      dependencies: lowest
-                      db: pgsql
                     - php-version: '7.2'
                       dependencies: lowest
                       db: sqlite


.github/workflows/test-application.yaml
```diff
@@ -41,6 +42,15 @@ jobs:
...
                       db: pgsql
+                    - php-version: '7.2'
+                      dependencies: lowest
+                      db: sqlite
+                    - php-version: '7.2'
+                      dependencies: lowest
+                      db: mysql
+                    - php-version: '7.2'
+                      dependencies: lowest
+                      db: pgsql
```
dbu, line 53: having one lowest build is enough, please remove the 7.3 lowest entries.

composer.json
```diff
@@ -15,20 +15,23 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.2 || ^8.0",
         "ext-pdo": "*",
         "doctrine/dbal": "^2.11 || ^3.0",
+        "doctrine/cache": "^1.12",
```
dbu, line 21: i am a bit unhappy with this, i'd prefer to keep being able to install jackalope along with doctrine/cache 2 - otherwise we create problematic combinations of requirements.

how is the problem related to PHP 7.2? or is it a general problem of the jackalope-doctrine-dbal 1.x branch?

dbu: ah ok, just found that doctrine/cache 2 only has interfaces but not the implementations. and the 1.x branch fails. lets do this then.

PR #390, opened 2021-11-03
Improve performance when deleting a property

wachterjohannes:
I have found another performance problem when we have big nodes (10000+ properties). Inside of the function `Client::deleteProperty` we spend most of the time when updates such nodes (with write, delete operations). 

<hr>

# Before

Spend in Method: 21.7s

Profile https://blackfire.io/profiles/28b1016c-b67b-4f2c-a0cc-9dca6349a963/graph

image

# After

Spend in Method: 528ms

Profile https://blackfire.io/profiles/40b5060b-1888-49b0-8f5e-08d9c1cb12ac/graph

image

<hr>

To avoid iterating over all the properties this PR uses XPath to load only the needed property.

As i have not changed anything at the logic there is no change in the tests required.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -1843,33 +1843,32 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
 
         $dom = new DOMDocument('1.0', 'UTF-8');
         $dom->loadXML($xml);
+        $xpath = new DomXpath($dom);
 
         $found = false;
         $propertyName = PathHelper::getNodeName($path);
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@dbu what do you think about that change?

dbu (approved):
great job!

i would recommend to challenge your data modelling, as 10k properties will likely create problems in many places. but even for 30 or 50 properties, i imagine this approach to speed up deletion quite a bit as we don't naively loop over all properties.

tests are green, so i think we can safely merge this.

dbu:
https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.7.3

wachterjohannes:
@dbu thanks for the fast response and also the release :)

we are currently debugging performance issues for several clients and tackle often the problem with a lot of properties. Improving that step by step makes it better for everyone 😂  but yeah the soft border of 10000 property is known but sometimes hard to not break through it.

Thanks again

PR #394, opened 2021-12-13
Fix postgresql errors in QOMWalker

alexander-schranz:


tests/Jackalope/Transport/DoctrineDBAL/ClientTest.php
```diff
@@ -537,6 +538,9 @@ public function provideOrder()
                 'val"ue',
                 'val"ue DESC',
                 ['three', 'two', 'one'],
+                [MySQLPlatform::class]
+                // see https://stackoverflow.com/questions/70339679/use-extractvalue-against-correctly-escaped-xml-attribute-value-in-mysql
+                // currently mysql does not escape 'val"ue' the same was as "val&quot;ue" so the test fails
```
alexander-schranz, line 543: Not sure how we could support this type of queries also in MySQL but as they were not supported yet I did just skip the test here.

dbu: to understand correctly: before, it happened to work on mysql because we did not correctly escape? and now that we do escape, we see that mysql issue?

dbu (approved):

Issue #173, opened 2014-01-29
Database columns/results are case-sensitive.  


tlorens:
We're on an oracle database and all columns are upper case.  There's code in Client.php that iterates through arrays using lower-cased keys.  This obviously blows up  when results are upper-cased.  I've patched the code with a simple call to :  $row =array_change_key_case($row, CASE_LOWER); in a few spots.   You make the call on a better, more practical solution.

jjwdesign:
@dbu

I know this is a very old thread, but I'm going to look at making a PR for this change soon. I've recently forked this project and will make a few fixes for Oracle. The majority of the problems we are facing are case related.

I tried to add some middleware to our document manager factory class, but it only helps with certain issues.

```        $dbConfig = new ConfigurationDBAL;
        $dbConfig->setMiddlewares([
            new Middleware(ConnectionPortability::PORTABILITY_FIX_CASE, ColumnCase::LOWER)
        ]);
        // create a doctrine dbal connection
        $dbConn = DriverManager::getConnection($dbParams, $dbConfig);
```
How long do you think it would take to review a PR for this update and get the changes included? Maybe in a Dev branch we can test with? We are doing a huge code migration to Laminas and PHP 8+. As usual, we're on a strict timeline and a bit behind.

dbu:
hi, old indeed... but still valid as oracle does not go away :-D 

i am glad if you take this up. i would expect that if we make all places that construct queries respect the lower/upper case defined in [the schema](https://github.com/jackalope/jackalope-doctrine-dbal/blob/baeb69776f60a64a882513566ec727854968bbbf/src/Jackalope/Transport/DoctrineDBAL/RepositorySchema.php), we would not need any middleware to make this work, right? or is it more complicated than that?

in terms of BC, i am a bit torn between adjusting the schema case to match the code or the code to match whats in the schema. but afaik for all officially *currently* supported databases the case does not matter, so i'd rather keep the schema the same and adjust the places where we query. (if somebody edited their oracle schema after setting up jackalope to match the case in the code but did not contribute back the fixing, they will have to revert the schema to what it is supposed to be...)

for the process: i don't have any longer holidays planned, so i should be able to react relatively quickly. however, to do the pull request you will fork this repository anyways, and you can make composer use your fork until i merged and tagged a release. set up your fork as a repository in composer.json (and then adjust the require part to say `jackalope-doctrine-dbal: dev-your-branch-name`) and update dependencies with that.

```
[...]

PR #399, opened 2022-07-29
DA-47795 Correcting Oracle issues with row array key name casings.

jjwdesign:
Correcting Oracle issues with column name casing. Applied several array_change_key_case() function calls to lower case the data array keys. For more information see issue: https://github.com/jackalope/jackalope-doctrine-dbal/issues/173

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -634,7 +634,8 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         $resultSetUuids = [];
 
         // first iterate and build up an array of all the UUIDs in the result set
-        foreach ($rows as $row) {
+        foreach ($rows as &$row) {
+            $row = array_change_key_case($row, CASE_LOWER);
             $resultSetUuids[$row['identifier']] = $row['path'];
         }
 
@@ -1296,6 +1297,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         $paths = [];
 
         foreach ($rows as $row) {
+            $row = array_change_key_case($row, CASE_LOWER);
             $this->nodeIdentifiers[$row['path']] = $row['identifier'];
             $data[$row['path']] = $this->xmlToProps($row['props']);
             $data[$row['path']]->{'jcr:primaryType'} = $row['type'];
@@ -1321,6 +1323,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         }
 
         foreach ($childrenRows as $child) {
+            $child = array_change_key_case($child, CASE_LOWER);
             $childName = explode('/', $child['path']);
             $childName = end($childName);
             if (!isset($data[$child['parent']]->{$childName})) {
@@ -1464,6 +1467,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         if (!$row) {
             throw new ItemNotFoundException("Item $uuid not found in workspace ".$this->workspaceName);
         }
+        $row = array_change_key_case($row, CASE_LOWER);
 
         $path = $row['path'];
         $data = $this->getNodeData($row);
@@ -1801,6 +1805,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         $values = $ids = [];
         $srcAbsPathPattern = '/^'.preg_quote($srcAbsPath, '/').'/';
         while ($row = $stmt->fetchAssociative()) {
+            $row = array_change_key_case($row, CASE_LOWER);
             $values['id'.$i] = $row['id'];
             $values['path'.$i] = preg_replace($srcAbsPathPattern, $dstAbsPath, $row['path'], 1);
             $values['parent'.$i] = PathHelper::getParentPath($values['path'.$i]);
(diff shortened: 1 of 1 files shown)


dbu (commented):
ah, ok. so the problem is that oracle messes up the case and then our code gets confused?

i wonder if we could decorate the connection with a wrapper instead of sprinkling the array_change_key_case all over the code. that way we could guarantee that we don't forget it somewhere, and we could use a different decorator depending on the platform so we only lowercase oracle and not the others. would be a bit tricky because we also would need a wrapper for the statement. lets try to get it working like this, and i will see if i can refactor afterwards.

i did some googling but without success, trying to find out if it would be possible to run our phpunit test suite with an oracle database. that would be helpful to know if we covered everything or if there are still problems. do you know if that is possible without buying oracle licenses or whatnot?

dbu:
about the test failures: i will have a look in a separate branch, seems to be code rot and not related to your changes.

dbu:
and one more thing: the master branch represents the unreleased 2.0. i would prefer if you would do the change against the 1.x branch so we can quickly release it once it is fine. there is no BC break involved so we can release it as a new minor version.

dbu:
i fixed the static tests (phpstan and cs-fixer). the phpunit failures seem to be related to the changes in this branch. there seems to be some side effect from lowercasing everything. can you investigate that please?

jjwdesign:
> ah, ok. so the problem is that oracle messes up the case and then our code gets confused?

Yeah, basically.

> i wonder if we could decorate the connection with a wrapper instead of sprinkling the array_change_key_case all over the code. that way we could guarantee that we don't forget it somewhere, and we could use a different decorator depending on the platform so we only lowercase oracle and not the others. would be a bit tricky because we also would need a wrapper for the statement. lets try to get it working like this, and i will see if i can refactor afterwards.

I thought I had applied a middleware on the dbal connection. I need to look back at this. Perhaps the code change was not merged or my co-worker was working with some old project code. Here's what my PHPCR nodes factory class looks like. It sets up the document manager.

```
<?php
namespace MyWhateverApp\Factory;

use Psr\Container\ContainerInterface,
    Laminas\ServiceManager\Factory\FactoryInterface,
    Doctrine\DBAL\DriverManager,
    Doctrine\DBAL\Portability\Middleware,
    Doctrine\DBAL\Configuration as ConfigurationDBAL,
    Doctrine\DBAL\Portability\Connection as ConnectionPortability,
    Doctrine\DBAL\ColumnCase,
    Jackalope\RepositoryFactoryDoctrineDBAL,
    PHPCR\SimpleCredentials,
    Doctrine\ODM\PHPCR\Configuration as ConfigurationPHPCR,
    Doctrine\Common\Annotations\AnnotationReader,
    Doctrine\ODM\PHPCR\Mapping\Driver\AnnotationDriver,
    Doctrine\ODM\PHPCR\DocumentManager;


class DoctrinePHPCRFactory implements FactoryInterface
{

    public function __invoke(ContainerInterface $container, $requestedName, array $options = null): mixed
    {

        $config = $container->get('config')['doctrine-phpcr-odm'];

        // Transport implementation
        $dbParams = [
            'driver' => $config['driver'],
            'user' => $config['dbuser'],
            'password' => $config['dbpass'],
            'dbname' => $config['dbname'],
            'host' => $config['dbhost'],
            'port' => $config['dbport'],
            'service' => $config['service'] ?? null
        ];
        $dbConfig = new ConfigurationDBAL;
        $dbConfig->setMiddlewares([
            new Middleware(ConnectionPortability::PORTABILITY_FIX_CASE, ColumnCase::LOWER)
        ]);
        // create a doctrine dbal connection
        $dbConn = DriverManager::getConnection($dbParams, $dbConfig);

        // Get jackalope repository
        $parameters = array('jackalope.doctrine_dbal_connection' => $dbConn);
        $dbal = new RepositoryFactoryDoctrineDBAL;
        $repository = $dbal->getRepository($parameters);

        $credentials = new SimpleCredentials($config['repository-username'], $config['repository-password']);
        $session = $repository->login($credentials, $config['repository-workspace']);

        // Prepare the doctrine configuration and annotations reader
        $configObj = new ConfigurationPHPCR();
        $reader = new AnnotationReader();

        // Autoloaded annotations directories
        $annotationPaths = array_unique(array_values($config['annotation-entity-path']));

        // Set Mapping Driver
        $driver = new AnnotationDriver($reader, $annotationPaths);

        $configObj->setMetadataDriverImpl($driver, false);
        $configObj->setAutoGenerateProxyClasses(true);
        $configObj->setProxyNamespace($config['proxy-namespace']);
        $configObj->setProxyDir($config['proxy-directory']);

        return DocumentManager::create($session, $configObj);
    }

}
```

dbu:
i understood that this middleware does not fix all issues? my thought was to create a decorator class for the return value of getConnection (and the statement class) that only implements the few methods actually used in the code. that way we could prevent accidentally not lower cased things - and could apply this only when the backend is oracle.

jjwdesign:
> i understood that this middleware does not fix all issues?

After further investigation, we had multiple document manager sessions being created. This caused some confusion with the middleware code. The middleware code shown above does in fact work. The other sessions that we had did not use the middleware to lowercase the field names. We can close this PR. Thank you.

dbu:
oh, great to hear!

we do not instantiate the connection ourselves in RepositoryFactoryDoctrineDBAL. nonetheless, we could check if we are on oracle and add the middleware if its not already in. does #401 make sense? may i ask you to try this out with your oracle installation to see if you can then get rid of the custom logic you currently have?

dbu:
superseeded in #401 , thanks a lot for explaining what was missing to work with oracle.

PR #408, opened 2023-01-01
Allow Doctrine Cache 2

derrabus:
Doctrine Cache is not maintained anymore. Version 2 is a transitional package with PSR-6 wrappers.

This PR allows the installation of v2 and uses Symfony Cache as a cache implementation for tests. The one problem I could not solve easily is this fallback here:

https://github.com/jackalope/jackalope-doctrine-dbal/blob/7a17d5955c52cd3a80b9f6e0b98da360f838917d/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php#L41

I chose to throw an exception if the `meta` cache has not been configured and `ArrayCache` is not available.

In the long run, this package should probably transition to PSR-6 and abandon Doctrine Cache completely.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "php": "^7.2 || ^8.0",
         "ext-pdo": "*",
         "doctrine/dbal": "^2.11 || ^3.0",
-        "doctrine/cache": "^1.12",
+        "doctrine/cache": "^1.12 || ^2",
         "phpcr/phpcr": "~2.1.5",
         "phpcr/phpcr-utils": "^1.5.0",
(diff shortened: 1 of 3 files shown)


dbu (approved):
sorry, that went under. seems good to me. i am currently preparing jackalope 2.0

does it make sense to drop support for doctrine/cache 1 completely? when we are allowed to break BC, what is the best way to handle the cache? should we completely switch to PSR cache?

dbu:
https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.9.0

dbu:
i created #411 to discuss about caching in jackalope 2.0

PR #409, opened 2023-02-20
Fix query syntax to work on MySQL with ANSI_QUOTES enabled

hanzi:
If MySQL is configured with SQL mode `ANSI_QUOTES` (or `ANSI`) the double quotes character is considered to quote an identifier name (such as table name, column name, etc.) rather than a literal string.

There is one place in the code base (which this PR aims to fix) where double quotes are being generated as part of the query, and running such a query on a MySQL server with `ANSI_QUOTES` enabled leads to an SQL error because the value of `$literalOperand` is being treated as a column name.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Query/QOMWalker.php
@@ -758,7 +758,7 @@ class QOMWalker
 
         if ($this->platform instanceof MySQLPlatform && '=' === $operator) {
             return sprintf(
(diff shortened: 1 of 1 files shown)


dbu (approved):
just red up a bit on the topic and it generally seems that single quotes are the correct thing to do in SQL.

thanks a lot for the fix!

(btw "interesting" how we already had single quotes for the xpath expression, but not the other strings...)

dbu:
https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.8.1

PR #416, opened 2023-02-27
do not attempt to create CachedClient if no meta & ArrayCache not available

dbu:
this gets us closest to the previous behaviour but not unnecessary errors.

@alexander-schranz does that help with your worry about BC? i hope you are using the repository factory in sulu too?


alexander-schranz:
@dbu Looks good for me!

PR #421, opened 2023-03-23
Taking advantage of batching property deletes when updating nodes

mamazu:
## Why
Currently the implementation for deleting a list of properties is to iterate over them and treat every delete of a property separately. This is a real determent to the performance of PHPCR, especially with bigger nodes.

Example: I want do delete three properties from the same node in the database. **For every property** we want to delete this is what it does:
* Check if the user is logged in.
* Load the node (SELECT query)
* Parse the document for it
* Remove property from the XML
* Update the XML in the database (UPDATE query)

> Which in turn means if you want to delete two properties from the same node, then you have to query the database for the node twice and also parse the document twice.

## The solution
Instead of defining the batch delete as a list of single deletes why not have the single delete make a batch delete with one element.

This pull request extracts out the logic for deleting a property from a node and extends it that you can delete a list of properties on one node as well.

To illustrate this with the same example: This would be the steps the program takes:
* Check login status at the start of the delete process
* Load the node once (1 SELECT query)
* Parse the document for it once
* Remove all properties from the XML in one go
* Update the XML in the database (1 UPDATE query)

> This massively reduces the amount of parsing the document and the amount of reads and writes to the database.

## Drawbacks
The new "batching functionality" first creates a map of properties and nodes to delete. Creating this map takes more memory than the brute force approach before.

## Others
Maybe we could also backport this speed up the 1.x branch so that project which are currently using the jackalope doctrine adapter also get the speed up.

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -1548,7 +1548,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
             $nodeId = $this->getSystemIdForNode($nodePath);
             if (!$nodeId) {
                 // no we really don't know that path
-                throw new ItemNotFoundException('No item found at '.$path);
+                throw new ItemNotFoundException('No item found at '.$nodePath);
             }
             if (!array_key_exists($nodeId, $nodesById)) {
                 $nodesById[$nodeId] = [];
@@ -1587,10 +1587,10 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
     }
 
     /**
-     * Removes a list of properties from a given node
+     * Removes a list of properties from a given node.
      *
      * @param string|int $nodeId
-     * @param array<string> $paths  Path belonging to that node that should be deleted
+     * @param array<string> $paths Path belonging to that node that should be deleted
      */
     private function removePropertiesFromNode(string|int $nodeId, array $paths): void
     {
@@ -1602,7 +1602,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         $xpath = new \DOMXPath($dom);
 
         $found = false;
-        foreach($paths as $path) {
+        foreach ($paths as $path) {
             $propertyName = PathHelper::getNodeName($path);
             $tablesToDeleteReferencesFrom = [];
             foreach ($xpath->query(sprintf('//*[@sv:name="%s"]', $propertyName)) as $propertyNode) {
@@ -1615,6 +1615,9 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
                 if (in_array($type, ['reference', 'weakreference'])) {
                     $tablesToDeleteReferencesFrom[] = $this->referenceTables['reference' === $type ? PropertyType::REFERENCE : PropertyType::WEAKREFERENCE];
                 }
+
+                // Doing the XML removal
+                $propertyNode->parentNode->removeChild($propertyNode);
             }
 
             if (!$found) {
@@ -1623,21 +1626,18 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
             }
 
             // Deleting the references
-            foreach(array_unique($tablesToDeleteReferencesFrom) as $table) {
+            foreach (array_unique($tablesToDeleteReferencesFrom) as $table) {
                 try {
                     $query = "DELETE FROM $table WHERE source_id = ? AND source_property_name = ?";
                     $this->getConnection()->executeQuery($query, [$nodeId, $propertyName]);
                 } catch (DBALException $e) {
                     throw new RepositoryException(
-                        'Unexpected exception while cleaning up deleted nodes',
+                        "Can not delete references for property $propertyName from `$table`",
                         $e->getCode(),
                         $e
                     );
                 }
             }
-
-            // Doing the XML removal
-            $propertyNode->parentNode->removeChild($propertyNode);
         }
 
         $xml = $dom->saveXML();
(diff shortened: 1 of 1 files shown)


mamazu:
@dbu Could you have a look at this? This could be a big performance improvement for everyone trying to delete properties.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1576,33 +1602,42 @@ private function deleteProperty(string $path): void
...
-                        $query = "DELETE FROM $table WHERE source_id = ? AND source_property_name = ?";
-                        $this->getConnection()->executeUpdate($query, [$nodeId, $propertyName]);
-                    } catch (DBALException $e) {
-                        throw new RepositoryException(
-                            'Unexpected exception while cleaning up deleted nodes',
-                            $e->getCode(),
-                            $e
-                        );
-                    }
+                    $tablesToDeleteReferencesFrom[] = $this->referenceTables['reference' === $type ? PropertyType::REFERENCE : PropertyType::WEAKREFERENCE];
```
dbu, line 1617: is this an additional optimization to avoid running unncessary deletions in n-m relationships?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1541,8 +1541,24 @@ public function deleteProperties(array $operations): void
     {
         $this->assertLoggedIn();
 
+        // Building a table describing which properties need to be deleted from which nodes so that we only have to parse contents once
```
dbu:
```suggestion
        // Building a table describing which properties need to be deleted from which nodes so that we only have to parse contents of each node once
```

dbu (approved):
thanks, interesting example of trading additional code complexity for performance.

the changes make sense, and if many properties are deleted, i can imagine the performance impact can be considerable.

i asked one question and suggest a minor documentation change, otherwise looks good to me.

dbu:
it seems that phpstan sees some unhandled edge cases. can you please cover those? and run the cs fixer?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1567,7 +1583,17 @@ private function deleteProperty(string $path): void
...
+        $this->removePropertiesFromNode($nodeId, [$path]);
+    }
 
+    /**
+     * Removes a list of properties from a given node
+     *
+     * @param string|int $nodeId
+     * @param array<string> $paths  Path belonging to that node that should be deleted
+     */
+    private function removePropertiesFromNode(string|int $nodeId, array $paths): void
```
mamazu, line 1594: php 8.1? Is that okay or do we need backporting to older versions?

src/Jackalope/Transport/DoctrineDBAL/Client.php
mamazu: I wasn't really sure about this one. But the reasoning is the following in the old code it would fire a delete for every instance of the property being found by the xpath selector
```php
$propertyName = PathHelper::getNodeName($path);
$tablesToDeleteReferencesFrom = [];
foreach ($xpath->query(sprintf('//*[@sv:name="%s"]', $propertyName)) as $propertyNode) {
```

However if the query is depending on the `$nodeId` and the `$propertyName` neither of them change in the loop. So running only at most one delete query for a given property seems better.

I don't know if you could simplify it even further and just accumulate references that need to be deleted per node and then delete all of them at the end. (Before flushing the new XML content). It's always a tadeoff between memory and speed.

mamazu:
I fixed the errors phpstan reported, they were actually bugs. But now it should be clean maybe cs is still broken 'cause I tried to fix it manually.

mamazu:
Furthermore: Should I rebase this PR to 1.x and then upmerge it? Because I think projects like sulu/sulu could really benefit from that right now.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: makes sense to me. i think your improvement is good enough.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: the upcoming 2.0 version of jackalope is targetting 8.0

8.0 is already old again, but lets stick with it for now. meaning you have to remove the type declaration in the signature and we rely on the phpdoc typehint.

dbu: the 8.0 tests are happy. are you sure this is not allowed with php 8.0?

dbu:
> Furthermore: Should I rebase this PR to 1.x and then upmerge it? Because I think projects like sulu/sulu could really benefit from that right now.

is deleting several properties a pattern that comes up a lot in sulu? i would honestly prefer to have the change in the new major version, to favor stability over performance improvements - there is the off-chance that there is some side effect in this refactoring.
and 2.0 should be released soon :tm: i already tagged the first beta release of jackalope/jackalope

dbu:
looks good to me :+1: 

can you please fix the remaining CS complaints and add an entry in the changelog for 2.0?

src/Jackalope/Transport/DoctrineDBAL/Client.php
mamazu: Yeah you are right. This is a php8.0 feature.

dbu:
thanks a lot!

mamazu:
> is deleting several properties a pattern that comes up a lot in Sulu?

It doesn't occur at all in Sulu yet. I could speculate that the previously bad performance might be a reason, but I am working on commands to clean the phpcr repository in Sulu and there a well performing delete is necessary for this.

But I guess 2.0 could be also a good idea in case any changes are BC breaks.

dbu:
there should be no BC breaks, but it is a bigger change in the logic that i feel more comfortable as releasing with 2.0.0. consumers are expected to test new major versions more thoroughly than minor releases.

PR #427, opened 2023-10-02
Adding return types to commands

mamazu:
Not specifying that the `configure` function is `void` is deprecated.


dbu (approved):
i have missed this when cleaning up the code for version 2. thanks a lot for spotting!

PR #429, opened 2023-11-21
Fix cache get

chirimoya:


dbu:
thanks a lot, fixed in #430 and released as https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/1.10.1

PR #432, opened 2023-11-23
Improve delete properties performance by replace DOMDocument with xml_parse

alexander-schranz:
See also: https://github.com/jackalope/jackalope-doctrine-dbal/pull/423

----

Prototype Repository: https://github.com/alexander-schranz/jackalope-xml-delete-properties

# Jackalope Doctrine DBAL Analyse deletion of properties

The benchmark requires 2 files:

 - `var/props.xml` - the xml of the node where we want to remove properties (SELECT props FROM phpcr_nodes WHERE identifier = ?)
 - `var/props.csv` - the props names which we want remove from the given xml (per line one property name)

## Different commands:

```bash
php src/remover.php legacy
php src/remover.php single_dom_document
php src/remover.php single_dom_query
php src/remover.php single_dom_query_chunk
php src/remover.php xml_parse
```

### ~70000 properties (~12.5MB) remove ~1700 props

Run on a MacBook Pro (16", 2021) Apple M1 Pro 32 GB:

| Command                | Time   | Memory |
|------------------------|--------|--------|
| legacy                 | 6m 29s | 48 MB  |
| single_dom_document    | 2m 25s | 35 MB  |
| single_dom_query       | 1m 30s | 37 MB  |
| single_dom_query_chunk | 1m 29s | 35 MB  |
| xml_parse              | ~1s    | 35 MB  |

`legacy`: is the `1.9.0` version: https://github.com/jackalope/jackalope-doctrine-dbal/blob/f7b286f388e0d3a42497c29e597756d6e346fea5/src/Jackalope/Transport/DoctrineDBAL/Client.php#L1804
`single_dom_document`: should represent the state of `2.0.0-beta2` version after: https://github.com/jackalope/jackalope-doctrine-dbal/pull/423/files

**Blackfire**

I did not use Blackfire for benchmarking as it did show in past benchmark where `xml_parse`
can not be good profiled as having a lot of callback method being called via `xml_set_element_handler` and `xml_set_character_data_handler`.
So profiling takes more time as processing things as Blackfire need to log every method call.
Instead I depend on classic benchmarking via time() and memory_get_peak_usage(true) measures.

#### A: Group Properties

The most important thing is that we remove all properties at once instead of calling `saveXML` after each property removal.

For this we mostly would require first group all `deleteProperties` by its `node`:

```php
function groupByNode($deletePropertyPaths): array {
    $grouped = [];
    foreach ($deletePropertyPaths as $path) {
        $nodePath = PathHelper::getParentPath($path);
        $propertyName = PathHelper::getNodeName($path);

        $grouped[$nodePath][] = $propertyName;
    }

    return $grouped;
}
```

Then we load the single `node` remove all the properties and save the xml once via `saveXML`.

#### B: Grouped Reference delete queries

The `queries` to remove references should also be grouped and best a single query be send to delete the references
instead of one query per reference.

The queries are currently ignored in the benchmark as it is focused on XML manipulation.

#### C: Replace DOMDocument with xml_parse

DOMDocument is bad for performance and should be avoided.
The `xml_parse` as it allows us to streamed reading the xml and skip the properties which we want to remove.
The [`XmlPropsRemover`](src/XmlPropsRemover.php) is an example how this could be done.

#### D: TODO

 - [x] xml_parse escape attributes name correctly
 - [x] xml_parse escape attributes value correctly
 - [x] xml_parse data content correctly
 - [x] xml_parse add queries to remove references
 - [x] xml_parse make sure the output is the same as for DOMDocument
    - [x] self closing tags
    - [x] escape data

~~currently there is a little difference in the 2 printed xmls:~~

```
-rw-r--r--   1 staff  staff  11940901 22 Nov 23:38 removed_legacy.xml
-rw-r--r--   1 staff  staff  12002548 22 Nov 23:39 removed_xml_parse.xml
```

Update `xml_parse` variant now has the same output as the previous DOMDocument version:

```
-rw-r--r--   1 staff  staff  11940901 22 Nov 23:33 removed_legacy.xml
-rw-r--r--   1 staff  staff  11940901 23 Nov 00:18 removed_xml_parse.xml
-rw-r--r--   1 staff  staff  12009559 22 Nov 23:45 removed_legacy_pretty.xml
-rw-r--r--   1 staff  staff  12009559 23 Nov 00:18 removed_xml_parse_pretty.xml
```

----

TODO:

 - [x] Check Sulu compatibility https://github.com/sulu/sulu/pull/7210

### DIFF
--- a/src/Jackalope/Transport/DoctrineDBAL/Client.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/Client.php
@@ -31,6 +31,7 @@ use Jackalope\Query\Query;
 use Jackalope\Transport\BaseTransport;
 use Jackalope\Transport\DoctrineDBAL\Query\QOMWalker;
 use Jackalope\Transport\DoctrineDBAL\XmlParser\XmlToPropsParser;
+use Jackalope\Transport\DoctrineDBAL\XmlPropsRemover\XmlPropsRemover;
 use Jackalope\Transport\MoveNodeOperation;
 use Jackalope\Transport\NodeTypeManagementInterface;
 use Jackalope\Transport\QueryInterface as QueryTransport;
@@ -1798,10 +1799,19 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
      */
     public function deleteProperties(array $operations)
     {
-        $this->assertLoggedIn();
-
+        $nodesByPath = [];
         foreach ($operations as $op) {
-            $this->deleteProperty($op->srcPath);
+            $nodePath = PathHelper::getParentPath($op->srcPath);
+            $propertyName = PathHelper::getNodeName($op->srcPath);
+            if (!array_key_exists($nodePath, $nodesByPath)) {
+                $nodesByPath[$nodePath] = [];
+            }
+            $nodesByPath[$nodePath][$propertyName] = $propertyName;
+        }
+
+        // Doing the actual removal
+        foreach ($nodesByPath as $nodePath => $propertiesToDelete) {
+            $this->removePropertiesFromNode($nodePath, $propertiesToDelete);
         }
 
         return true;
@@ -1827,53 +1837,47 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
      */
     protected function deleteProperty($path)
     {
-        $this->assertLoggedIn();
-
         $nodePath = PathHelper::getParentPath($path);
+        $propertyName = PathHelper::getNodeName($path);
+        $this->removePropertiesFromNode($nodePath, [$propertyName]);
+    }
+
+    /**
+     * Removes a list of properties from a given node.
+     *
+     * @param string $nodeId
+     * @param array<string> $paths Path belonging to that node that should be deleted
+     */
+    private function removePropertiesFromNode($nodePath, array $propertiesToDelete): void
+    {
+        $this->assertLoggedIn();
         $nodeId = $this->getSystemIdForNode($nodePath);
         if (!$nodeId) {
             // no we really don't know that path
-            throw new ItemNotFoundException("No item found at " . $path);
+            throw new ItemNotFoundException('No item found at ' . $nodePath);
         }
 
         $query = 'SELECT props FROM phpcr_nodes WHERE id = ?';
         $xml = $this->getConnection()->fetchOne($query, [$nodeId]);
 
-        $dom = new DOMDocument('1.0', 'UTF-8');
-        $dom->loadXML($xml);
-        $xpath = new DomXpath($dom);
+        $xmlPropsRemover = new XmlPropsRemover($xml, $propertiesToDelete);
+        [$xml, $references] = $xmlPropsRemover->removeProps();
 
-        $found = false;
-        $propertyName = PathHelper::getNodeName($path);
-        foreach ($xpath->query(sprintf('//*[@sv:name="%s"]', $propertyName)) as $propertyNode) {
-            $found = true;
-            // would be nice to have the property object to ask for type
-            // but its in state deleted, would mean lots of refactoring
-            if ($propertyNode->hasAttribute('sv:type')) {
-                $type = strtolower($propertyNode->getAttribute('sv:type'));
-                if (in_array($type, ['reference', 'weakreference'])) {
-                    $table = $this->referenceTables['reference' === $type ? PropertyType::REFERENCE : PropertyType::WEAKREFERENCE];
-                    try {
-                        $query = "DELETE FROM $table WHERE source_id = ? AND source_property_name = ?";
-                        $this->getConnection()->executeUpdate($query, [$nodeId, $propertyName]);
-                    } catch (DBALException $e) {
-                        throw new RepositoryException(
-                            'Unexpected exception while cleaning up deleted nodes',
-                            $e->getCode(),
-                            $e
-                        );
-                    }
+        foreach ($references as $type => $propertyNames) {
+            $table = $this->referenceTables['reference' === $type ? PropertyType::REFERENCE : PropertyType::WEAKREFERENCE];
+            foreach ($propertyNames as $propertyName) {
+                try {
+                    $query = "DELETE FROM $table WHERE source_id = ? AND source_property_name = ?";
+                    $this->getConnection()->executeUpdate($query, [$nodeId, $propertyName]);
+                } catch (DBALException $e) {
+                    throw new RepositoryException(
+                        'Unexpected exception while cleaning up deleted nodes',
+                        $e->getCode(),
+                        $e
+                    );
                 }
             }
-
-            $propertyNode->parentNode->removeChild($propertyNode);
         }
-
-        if (!$found) {
-            throw new ItemNotFoundException("Node $nodePath has no property $propertyName");
-        }
-
-        $xml = $dom->saveXML();
 
         $query = 'UPDATE phpcr_nodes SET props = ? WHERE id = ?';
         $params = [$xml, $nodeId];
@@ -1881,7 +1885,7 @@ class Client extends BaseTransport implements QueryTransport, WritingInterface, 
         try {
             $this->getConnection()->executeUpdate($query, $params);
         } catch (DBALException $e) {
-            throw new RepositoryException("Unexpected exception while updating properties of $path", $e->getCode(), $e);
+            throw new RepositoryException("Unexpected exception while updating properties of $nodePath", $e->getCode(), $e);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1827,61 +1838,55 @@ public function deletePropertyImmediately($path)
...
+
+    /**
+     * Removes a list of properties from a given node.
+     *
+     * @param string $nodeId
+     * @param array<string> $paths Path belonging to that node that should be deleted
+     */
+    private function removePropertiesFromNode($nodePath, array $propertiesToDelete): void
+    {
+        $this->assertLoggedIn();
```
mamazu: Why do we check if the user is logged in for every node? Wouldn't it be sufficient to check once on process start?

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1827,61 +1838,55 @@ public function deletePropertyImmediately($path)
...
         $nodePath = PathHelper::getParentPath($path);
+        $propertyName = PathHelper::getNodeName($path);
+        $this->removePropertiesFromNode($nodePath, [$propertyName]);
+    }
+
+    /**
+     * Removes a list of properties from a given node.
+     *
+     * @param string $nodeId
+     * @param array<string> $paths Path belonging to that node that should be deleted
```
mamazu: Phpdocs should match the parameters.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: i agree with mamazu, would move this back to the 2 entry point methods. if properties from several nodes are removed, we otherwise check unnecessary often

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,165 @@
...
+                    return;
+                }
+
+                return;
+            }
+        }
+
+        $tag = '<' . \strtolower($name);
+        foreach ($attrs as $key => $value) {
+            $tag .= ' ' . \strtolower($key) . '="' . $value . '"'; // TODO escaping
```
dbu: this still says TODO, need to change something? is there an xml parser method maybe that could do this savely?

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private function dataHandler($parser, $data): void
+    {
+        if ($this->skipCurrentTag) {
+            return;
+        }
+
+        if ($data !== '') {
+            $this->newXml .= $this->newStartTag; // none empty data means no self closing tag so render tag now
```
dbu:
```suggestion
            $this->newXml .= $this->newStartTag; // non-empty data means no self closing tag so render tag now
```

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,165 @@
...
+    private $references = [];
+
+    public function __construct(string $xml, array $propertyNames)
+    {
+        $this->xml = $xml;
+        $this->propertyNames = $propertyNames;
+    }
+
+    /**
+     * @return array{
```
dbu, line 48: i'd love the phpdoc to also explain what the fields are, not only the types

tests/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemoverTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace Jackalope\Transport\DoctrineDBAL\XmlPropsRemover;
+
+use Jackalope\Factory;
+use Jackalope\Test\TestCase;
+use Jackalope\Transport\DoctrineDBAL\XmlParser\XmlToPropsParser;
+use PHPCR\Util\ValueConverter;
+
+class XmlPropsRemoverTest extends TestCase
```
dbu, line 10: can you please add a test that requires escaping of name/value and makes sure that all works as expected too?

dbu:
that is some really impressive improvement, thanks for the analysis and the implementation.

tests/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemoverTest.php
alexander-schranz: @dbu yeah I will check the JCR specification for what is even possible as jcr property names. The jcr values are already escaped tested via `ampersand` property.

alexander-schranz:
Thx for the reviews.

@dbu we will next week do some more profiling on some projects and then I will continue here with some fixes. Pushed this already so we can test it on various projects via:

```json
    "repositories": [
        {
            "type": "vcs",
            "url": "git@github.com:alexander-schranz/jackalope-doctrine-dbal.git"
        }
    ],
    "require": {
    
        "jackalope/jackalope-doctrine-dbal": "dev-feature/improve-delete-properties-performance as 1.10.0",
        
    }
```

dbu:
i am about to tag the next beta for jackalope-doctrine-dbal 2.0

i think there is nothing preventing us from doing 2.1 very quickly afterwards, once this is ready. but that way we at least should soon have a stable release that installs with symfony 7 - and this change has no BC breaks, it is an optimization that can come as minor version. (unless you see something that you would want to BC break to make things easier  - if we need that tell me by monday so we can look into it before releasing 2.0.0 stable.)

alexander-schranz:
@dbu I will do again a test run against sulu jackolope 2 tomorrow. To release this as new minor for 1 and/or 2 I also see no problem as it is just changing internal logic.

alexander-schranz:
I did do some adjustements so escaping the property name and value behave the same as before and extend the test case so we are 100% sure the XML special chars are escaped the same way as before.

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
alexander-schranz: This is implemented now.

tests/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemoverTest.php
alexander-schranz: Additional tests are added for more xml special chars `<>&'"` and tested

alexander-schranz:
Applied the suggestions to the pull request, so it is ready now :) Also did again more testing with special chars and the result is 100% the same for [old and new implementations](https://github.com/alexander-schranz/jackalope-xml-delete-properties):

Bildschirmfoto 2023-12-11 um 09 53 37

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if ($name === 'SV:PROPERTY') {
+            $svName = $attrs['SV:NAME'];
+
+            if (\in_array($svName, $this->propertyNames)) {
+                $this->skipCurrentTag = true;
+
+                if (\in_array($svName, $this->propertyNames)) {
+                    $this->skipCurrentTag = true;
```
mamazu: This looks like duplicated logic. Maybe some merging tool messed this up? Or we should add a comment why this is necessary.

alexander-schranz: good catch, removed!

mamazu (approved):

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1827,61 +1839,55 @@ public function deletePropertyImmediately($path)
      */
     protected function deleteProperty($path)
     {
+        $nodePath = PathHelper::getParentPath($path);
+        $propertyName = PathHelper::getNodeName($path);
         $this->assertLoggedIn();
```
dbu: can we keep the assertion at the top, for readability?

dbu (approved):
thanks a lot! the tests you added increase my confidence that we don't have regressions.

one last nitpick, and can you please add a changelog entry explaining what the change means for the user?

src/Jackalope/Transport/DoctrineDBAL/Client.php
alexander-schranz: updated :)

dbu:
released in 1.11.0

PR #434, opened 2024-01-06
1 to 2

dbu:


src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,156 @@
+<?php
+
+namespace Jackalope\Transport\DoctrineDBAL\XmlPropsRemover;
+
+/**
+ * @internal
+ */
+class XmlPropsRemover
```
dbu, line 10: modernized for PHP 8. but the problem seems not to come from this file - i copied the version from 1.x into my local installation and ran the test and it still fails.

dbu:
@alexander-schranz i am trying to merge #432 upstream but see a failure. the DeleteCascadeTest is new in version 2, did not exist in version 1. (the reason sqlite does not fail is that foreign keys are not supported and therefore the test is skipped).

do we change how we update the database? though its suprising to me that cascade delete would no longer work because of a different query.

src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1578,97 +1576,46 @@ public function deletePropertyImmediately(string $path): void
...
-                $type = strtolower($propertyNode->getAttribute('sv:type'));
-                switch ($type) {
-                    case 'reference':
-                        $tablesToDeleteReferencesFrom[] = $this->referenceTables[PropertyType::REFERENCE];
-                        break;
-                    case 'weakreference':
-                        $tablesToDeleteReferencesFrom[] = $this->referenceTables[PropertyType::WEAKREFERENCE];
-                        break;
-                    case 'binary':
-                        $deleteBinary = true;
```
dbu, line 1628: ah, this is a new feature in 2.x and i did not realize when solving conflicts. we need to figure out how to keep this working

dbu: i think this means XmlPropsRemover needs to also return the list of binary properties, like it returns the reference properties.

dbu: sorry, :duck: 

i think i solved it

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
```diff
@@ -0,0 +1,170 @@
...
+        if ($this->skipCurrentTag) {
+            return;
+        }
+
+        if ('SV:PROPERTY' === $name) {
+            $svName = $attrs['SV:NAME'];
+
+            if (\in_array((string)$svName, $this->propertyNames, true)) {
+                $this->skipCurrentTag = true;
+                $svType = strtolower($attrs['SV:TYPE']);
```
dbu: with the binary removal test, the type was `Binary` and not `binary`. the node type definition have capitalized names for all types, but i guess it already worked for references (or is that not covered by the tests?)

not sure what exactly is going on. if you have an idea @alexander-schranz we could clean it up, but otherwise it should work like this.

tests/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemoverTest.php
```diff
@@ -0,0 +1,98 @@
...
+	</sv:property>
+	<sv:property sv:name="jcr:uuid" sv:type="String" sv:multi-valued="0">
+		<sv:value length="36">0804f0c3-5250-4c2f-9d7e-7d0c99103026</sv:value>
+	</sv:property>
+	<sv:property sv:name="i18n:en-title" sv:type="String" sv:multi-valued="0">
+		<sv:value length="8">My Title</sv:value>
+	</sv:property>
+	<sv:property sv:name="ampersand" sv:type="String" sv:multi-valued="0"><sv:value length="13">foo &amp; bar&amp;baz</sv:value></sv:property>
+	<sv:property sv:name="äüö?ß&lt;&gt;''&quot;=&quot;test" sv:type="String" sv:multi-valued="0"><sv:value length="15">&lt;&gt;:&amp;|öäü"?"ß'='</sv:value></sv:property>
+	<sv:property sv:name="block_1_ref" sv:type="reference" sv:multi-valued="0">1922ec03-b5ed-40cf-856c-ecfb8eac12e2</sv:property>
```
dbu, line 31: ah, actually @alexander-schranz are you sure that this is correct? should it not be `Reference` starting with capital `R`?

we might need to backport the strtolower to 1.x too

src/Jackalope/Transport/DoctrineDBAL/Client.php
alexander-schranz: Looks the same way I would have solve it :)

src/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemover.php
alexander-schranz: the tests seems to use lower case part. But I copied that from another tests so not sure if it was already there false.

tests/Jackalope/Transport/DoctrineDBAL/XmlPropsRemover/XmlPropsRemoverTest.php
alexander-schranz: I think it would make sense to backport the strtolower.

dbu: this seems to have been a mess already before: https://github.com/jackalope/jackalope-doctrine-dbal/blob/1.x/src/Jackalope/Transport/DoctrineDBAL/Client.php#L746

i tried to find the system view xsd definition but the URI does not lead to a download link. and the whole jackrabbit source code neither contains the schema nor links to it with schemaLocation :-(

i will backport the fix and see if there are other places with invalid assumptions.

alexander-schranz: interesting then it does also make more sense to use strtolower here.

dbu: i now use the PHPCR PropertyType static methods to avoid case errors.

PR #435, opened 2024-01-08
fix reference deletion to not be case sensitive

dbu:
as found in https://github.com/jackalope/jackalope-doctrine-dbal/pull/434

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Changelog
 1.x
 ===
 
+1.11.1
+------
+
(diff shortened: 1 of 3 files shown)


dbu:
@alexander-schranz does that make sense? i think it actually is a regression of 1.11.0, we would no longer delete dangling references if the xml uses sv:type with uppercase name.

alexander-schranz (approved):

alexander-schranz:
Looks fine from my side, didn't have any effects on performance improvements as tested in: https://github.com/alexander-schranz/jackalope-xml-delete-properties

dbu:
thanks. i think there would be no visible effects of the problem, "just" data not being deleted from a table but not doing any harm.

PR #439, opened 2024-02-02
Add support for doctrine/dbal 4 version

### LINKED ISSUE #441: Add compatibility with doctrine/dbal 4

dbu:
continue #436 

fix #441

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -9,7 +9,7 @@ on:
       - '[0-9]+.[0-9]+.x'
 jobs:
   test:
-    name: 'PHP ${{ matrix.php-version }}, Database ${{ matrix.db }} ${{ matrix.dependencies }}, Composer ${{ matrix.composer-stability }}'
+    name: 'PHP ${{ matrix.php-version }}, Database ${{ matrix.db }} ${{ matrix.dependencies }}'
     runs-on: ubuntu-20.04
 
     env:
@@ -30,9 +30,6 @@ jobs:
           - sqlite
         dependencies:
           - highest
-        composer-stability:
-          - dev
-          - stable
         include:
           - php-version: '8.0'
             dependencies: lowest
@@ -75,10 +72,6 @@ jobs:
           php-version: ${{ matrix.php-version }}
           extensions: "pdo, pdo_sqlite, pdo_mysql, mysql, pdo_pgsql"
           tools: 'composer:v2'
-
-      - name: Set composer stability
-        if: ${{ matrix.composer-stability }}
-        run: composer config minimum-stability ${{ matrix.composer-stability }}
 
       - name: PHP 8.0 simple cache
         # Symfony 5 is not compatible with SimpleCache 3 but does not declare a conflict. Symfony 6 can not be installed on PHP 8.0.
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -9,7 +9,7 @@ on:
       - '[0-9]+.[0-9]+.x'
 jobs:
   test:
-    name: 'PHP ${{ matrix.php-version }}, Database ${{ matrix.db }} ${{ matrix.dependencies }}'
+    name: 'PHP ${{ matrix.php-version }}, Database ${{ matrix.db }} ${{ matrix.dependencies }}, Composer ${{ matrix.composer-stability }}'
     runs-on: ubuntu-20.04
 
     env:
@@ -30,6 +30,9 @@ jobs:
           - sqlite
         dependencies:
           - highest
+        composer-stability:
+          - dev
+          - stable
         include:
           - php-version: '8.0'
             dependencies: lowest
@@ -43,7 +46,7 @@ jobs:
 
     services:
       mysql:
-        image: mysql:5.7
+        image: mysql:8.3
         env:
           MYSQL_ROOT_PASSWORD: root
           MYSQL_DATABASE: phpcr_tests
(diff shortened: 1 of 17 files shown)


src/Jackalope/Transport/DoctrineDBAL/Client.php
```diff
@@ -1679,10 +1680,19 @@ private function moveNode(string $srcAbsPath, string $destAbsPath): void
             throw new PathNotFoundException("Parent of the destination path '".$destAbsPath."' has to exist.");
         }
 
-        $query = 'SELECT path, id FROM phpcr_nodes WHERE path LIKE ? OR path = ? AND workspace_name = ? '.$this->getConnection()
-                ->getDatabasePlatform()->getForUpdateSQL()
+        $sql = $this->getConnection()->createQueryBuilder()
+            ->select('path, id')
+            ->from('phpcr_nodes')
+            ->where('(path LIKE :path_prefix OR path = :path) AND workspace_name = :workspace')
+            ->forUpdate()
```
dbu: @derrabus did i understand your suggestion in https://github.com/jackalope/jackalope-doctrine-dbal/pull/436#discussion_r1470991804 wrong?

this breaks sqlite with "FOR UPDATE" not supported errors. or is that a bug in dbal and it is supposed to correctly decide if it should put FOR UPDATE into the statement?

.github/workflows/test-application.yaml
```diff
@@ -43,12 +46,15 @@ jobs:
...
-        image: mysql:5.7
+        image: mysql:8.3
         env:
           MYSQL_ROOT_PASSWORD: root
           MYSQL_DATABASE: phpcr_tests
         options: >-
           --health-cmd "mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=5
+        steps:
+          - name: start
+            run: docker run -d mysql:8.3 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
```
dbu: found this in https://github.com/orgs/community/discussions/25086 but seems to also no longer work.

after trying 10 other things. @derrabus by any chance would you know how i can tell github actions what charset to use? it seems to default to utf8mb3 (man i hate stupid mysql collations mess)

derrabus: Does it still work for MySQL 5.7 though? I mean, you don't have to upgrade to 8.3 in order to support DBAL 4. You can still do that in a follow-up.

Chris53897: @dbu i am not sure if it solves the probleme, but if i read the example you shared correct, the `steps` is maybe here correct?
https://github.com/jackalope/jackalope-doctrine-dbal/pull/440/files#diff-97081780ba6b676e25ca4729b13035dd4d3fd5f010dd76151028ae821707d3cdR69

But ci is not running on my fork, and runs need approvals

src/Jackalope/Transport/DoctrineDBAL/Client.php
derrabus: The purpose of that method is to pick the right syntax for the lock hint. However, if your database does not support this kind of locks, the method does not silently fall back to no locking at all.

The idea is that it should be a deliberate decision on your side to not lock.

```php
        $qb = $this->getConnection()->createQueryBuilder()
            ->select('path, id')
            ->from('phpcr_nodes')
            ->where('(path LIKE :path_prefix OR path = :path) AND workspace_name = :workspace')
        ;

        if (!$this->getConnection()->getDatabasePlatform instanceof SQLitePlatform) {
            $qb->forUpdate();
        }

        $sql = $qb->getSQL();
```

But you're not the first to run into this problem. Maybe we will implement a configurable fallback strategy on our side.

.github/workflows/test-application.yaml
dbu: if i switch back to old mysql, the default encoding indeed seems to be utf8 (which afaik is fragile and its recommended to use utf8mb4 instead).

i would love to switch to utf8mb4 by default for mysql to reflect best practices. and as 2.x is not yet released, it would be the right moment for such a BC break.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: this still happens for sqlite, now that reverting mysql version and encoding got us rid of the fatal errors.

.github/workflows/test-application.yaml
dbu: also, dbal 4 seems to incorrectly report latin1 as encoding, when in reality it should be utf8. the PHP 8.0 builds install dbal 3.8.1, the PHP 8.1-3 install dbal 4.

src/Jackalope/Transport/DoctrineDBAL/Client.php
dbu: okay, i see. in the context of the discussion (where we just got it from whichever platform was being used and got either nothing or the correct statement) i thought the query builder would be aware of the platform.

the unfortunate thing is that this makes the application think about specific implementations and have assumptions. it would fail with custom platform implementations or new implementations that also do not support the lock hint, or if some future sqlite version would support it...

thanks for the explanation, then i will hardcode the check for now.

derrabus: > i thought the query builder would be aware of the platform.

It is and this is why you get an exception. 🙃 

> the unfortunate thing is that this makes the application think about specific implementations and have assumptions. it would fail with custom platform implementations or new implementations that also do not support the lock hint, or if some future sqlite version would support it...

Yes, it would. I understand that this is a problem for you, hence the last paragraph of my previous comment. 🙂

PR #440, opened 2024-02-04
chore: test ci for mysql service

Chris53897:


.github/workflows/test-application.yaml
```diff
@@ -69,6 +66,9 @@ jobs:
           --health-cmd "pg_isready" --health-interval 10s --health-timeout 5s --health-retries 5
 
     steps:
+      - name: start
+        run: docker run -d mysql:8.3 --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
```
dbu, line 70: if i understand correctly, this starts another mysql instance but we keep using the one from services.

dbu:
i decided in #439 to require the encoding to be set, either with the `charset` param to dbal, the `defaultTableOptions`.`collate` or by calling `Client::setCaseSensitiveEncoding()`

thanks again for the help on this one.

Chris53897:
i did not actualy help much. It was just a quick try.
Thank you for working on the dbal 4 support

PR #447, opened 2024-05-03
Fix PSR16 compatibility by sanitizing Colon symbol

alexander-schranz:
See https://github.com/sulu/sulu/issues/7394

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -75,6 +75,10 @@ jobs:
                   extensions: "pdo, pdo_sqlite, pdo_mysql, mysql, pdo_pgsql"
                   tools: 'composer:v2'
 
+            - name: Require dependencies
+              if: ${{ matrix.php == '8.0' }}
+              run: composer require "psr/simple-cache:2.*" --no-update --dev
+
             - name: Install dependencies with Composer
               uses: ramsey/composer-install@v1
               with:
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -76,7 +76,7 @@ jobs:
                   tools: 'composer:v2'
 
             - name: Require dependencies
-              if: ${{ matrix.php == '8.0' }}
+              if: ${{ matrix.php-version == '8.0' }}
               run: composer require "psr/simple-cache:2.*" --no-update --dev
 
             - name: Install dependencies with Composer
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -77,7 +77,7 @@ jobs:
 
             - name: Require dependencies
               if: ${{ matrix.php-version == '8.0' }}
-              run: composer require "psr/simple-cache:2.*" --no-update --dev
+              run: composer require "psr/simple-cache:2.*" --no-update --dev # This can be removed if min version of symfony/cache is ^6.0
 
             - name: Install dependencies with Composer
               uses: ramsey/composer-install@v1
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
     },
     "require-dev": {
         "psr/log": "^1 || ^2 || ^3",
+        "psr/simple-cache": "^1.0 || ^2.0 || ^3.0",
         "phpcr/phpcr-api-tests": "2.1.22",
         "phpunit/phpunit": "8.5.21",
         "symfony/cache": "^5.4 || ^6.2"
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -51,7 +51,7 @@ class CachedClient extends Client
 
         $this->caches = $caches;
         $this->keySanitizer = static function ($cacheKey) {
-            return str_replace(' ', '_', $cacheKey);
+            return str_replace([' ', ':'], '_', $cacheKey);
         };
     }
 
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -51,7 +51,11 @@ class CachedClient extends Client
 
         $this->caches = $caches;
         $this->keySanitizer = static function ($cacheKey) {
-            return str_replace([' ', ':'], '_', $cacheKey);
+            return str_replace(
+                [' ', ':', '{', '}', '(', ')', '.', '@', '/', '"', '\\'],
+                ['_', '|', '[', ']', '[', ']', '|', 'a', '|', '\'', '|'],
+                $cacheKey
+            );
         };
     }
 
--- a/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
+++ b/src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
@@ -52,9 +52,9 @@ class CachedClient extends Client
         $this->caches = $caches;
         $this->keySanitizer = static function ($cacheKey) {
             return str_replace(
-                [' ', ':', '{', '}', '(', ')', '.', '@', '/', '"', '\\'],
-                ['_', '|', '[', ']', '[', ']', '|', 'a', '|', '\'', '|'],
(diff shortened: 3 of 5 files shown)


src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -51,7 +51,7 @@ public function __construct(FactoryInterface $factory, Connection $conn, array $
 
         $this->caches = $caches;
         $this->keySanitizer = static function ($cacheKey) {
-            return str_replace(' ', '_', $cacheKey);
+            return str_replace([' ', ':'], '_', $cacheKey);
```
alexander-schranz: @dbu do you want to use another symbol as `_` to replace `:`?

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -45,7 +45,7 @@ public function testDefaultKeySanitizer()
         $this->cacheMock
             ->method('fetch')
             ->with(self::callback(function ($arg) use (&$first) {
-                self::assertEquals($first ? 'nodetypes:_a:0:{}' : 'node_types', $arg);
+                self::assertEquals($first ? 'nodetypes__a_0_{}' : 'node_types', $arg);
```
alexander-schranz: If I look at the error message which @flogado posted here: https://github.com/sulu/sulu/issues/7394

I think we also run into issue with the `{}`?

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
```diff
@@ -31,7 +31,7 @@ public function testArrayObjectIsConvertedToArray()
     public function testCacheHit()
     {
         $cache = new \stdClass();
-        $this->cacheMock->method('fetch')->with('nodes:_/test,_tests')->willReturn($cache);
+        $this->cacheMock->method('fetch')->with('nodes__/test,_tests')->willReturn($cache);
```
alexander-schranz:
```suggestion
        $this->cacheMock->method('fetch')->with('nodes|_|test,_tests')->willReturn($cache);
```

tests/Jackalope/Transport/DoctrineDBAL/CachedClientTest.php
alexander-schranz:
```suggestion
                self::assertEquals($first ? 'nodetypes|_a|0|[]' : 'node_types', $arg);
```

alexander-schranz:
@dbu I'm not sure how we can keep the functional test for the CacheClient here correctly as the dependency graph seems hard to control what is compatible with the specific symfony cache.

dbu:
thanks for this. kinda weird how many characters are not allowed in the key. [psr-16](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-16-simple-cache.md) specifially forbids `{}()/\@:` and says at minimum `A-Z, a-z, 0-9, _, and .` must be supported. we should probably transliterate everything onto the minimal characeter set. we could use `urlencode` and then replace % with _. having an explicit list of allowed things is better than escaping specific things and then the next person comes with a node with :small_airplane: in its name and some cache not happy with that.

RE simple-cache version: that is indeed unfortunate. it seems symfony 5 had psr-16 only as [dev-dependency](https://github.com/symfony/cache/blob/5.4/composer.json#L39) and thats the reason why composer does not know this won't work.

as this is the legacy 1.x branch, i propose we specifically do a `composer require "psr/simple-cache:2.*"` in the test-application.yaml file if php-version is equal to 8.0.

not super elegant, but its a bit of an edge case and not the reponsiblity of jackalope-doctrine-dbal .

want to do that?

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -664,7 +668,13 @@ private function set(string $name, string $key, $value): void
     private function get(string $name, string $key)
     {
         if ($this->caches[$name] instanceof CacheInterface) {
-            return $this->caches[$name]->get($key);
+            $result = $this->caches[$name]->get($key);
```
ghost: `CachedClient::get` no longer seems to exist in `2.x`. How can this be solved in `2.x`?

alexander-schranz: @flogado I changed the logic to return now null instead of false and adopted the code. Can you give that a try if that would work for 2.0?

alexander-schranz:
@dbu urlencode does not catch `.` so we need to replace atleast that character manually. I pushed a version and adopted the testcase.

dbu:
> @dbu urlencode does not catch `.` so we need to replace atleast that character manually. I pushed a version and adopted the testcase.

right. according to https://www.php.net/manual/en/function.urlencode.php that should be the only mismatch, right?

alexander-schranz:
@dbu I think so.

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
```diff
@@ -259,7 +263,7 @@ public function getNode($path)
         $cacheKey = "nodes: $path, ".$this->workspaceName;
         $cacheKey = $this->sanitizeKey($cacheKey);
 
-        if (false !== ($result = $this->get('nodes', $cacheKey))) {
+        if (null !== ($result = $this->get('nodes', $cacheKey))) {
```
dbu, line 266: why does this change from false to null?

composer.json
```diff
@@ -30,6 +30,7 @@
     },
     "require-dev": {
         "psr/log": "^1 || ^2 || ^3",
+        "psr/simple-cache": "^1.0 || ^2.0 || ^3.0",
```
dbu, line 33: do we directly interact with simple-cache, or why do we need this here?

.github/workflows/test-application.yaml
```diff
@@ -75,6 +75,10 @@ jobs:
                   extensions: "pdo, pdo_sqlite, pdo_mysql, mysql, pdo_pgsql"
                   tools: 'composer:v2'
 
+            - name: Require dependencies
+              if: ${{ matrix.php-version == '8.0' }}
+              run: composer require "psr/simple-cache:2.*" --no-update --dev
```
dbu: can you please add a comment why we need this? (to prevent the conflict symfony 5 if i remember correctly)

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
alexander-schranz: because PSR16 return `null` not false when [there is no result](https://github.com/php-fig/simple-cache/blob/2d280c2aaa23a120f35d55cfde8581954a8e77fa/src/CacheInterface.php#L18).

The current code was for the doctrine/cache (returning false), instead of converting null (psr16) to false (doctrine) I go with converting false (doctrine) to null (psr16) in the `get` method as @flogado mentiont that `2.x` does not have the `::get` method and always us the PSR16 implementation which returns `null` and not false, so all checks for `false !==` needed to convert to `null !==`.

composer.json
alexander-schranz: The CachedClientFunctionalTest requires this else it did throw an exception that psr/simple-cache is not installed.

.github/workflows/test-application.yaml
alexander-schranz: comment added

dbu: thanks

src/Jackalope/Transport/DoctrineDBAL/CachedClient.php
dbu: ah, now i get it. thanks, yeah better do it this way around.

composer.json
dbu: ah, and otherwise it uses the deprecated doctrine cache. makes sense.

in 2.x, this is even a full requirement: https://github.com/jackalope/jackalope-doctrine-dbal/blob/2.x/composer.json#L26 not dev only

dbu:
thanks a lot! squashed and added changelog entries in #448

Issue #450, opened 2024-07-09
Error inside the transport layer: Cache key 'nodes by uuid: 9ec4d67b-74e2-428a-a824-8f927e37749d, default' contains reserved characters '{}()/@:'.


Johann0290:
Hi together. Jackalope appears to have a bug in the production environment. This definitely occurs also in the version 2.0.1. Only in production, the function ClearNodeCache in CachedClient.php is called. In line 105, the cacheKey is set again without subsequently calling the sanitizeKey function. This leads to the error "Error inside the transport layer: Cache key 'nodes by uuid: 9ec4d67b-74e2-428a-a824-8f927e37749d, default' contains reserved characters '{}()/@:'.". The assigned string contains a ":".

dbu:
thanks for the clear report. indeed that place seems to have been missed when adding the key sanitize logic. fixed in #451 and tagged as https://github.com/jackalope/jackalope-doctrine-dbal/releases/tag/2.0.2

Johann0290:
Gladly. I also appreciate the quick fix.

PR #454, opened 2025-07-22
add missing cache key sanitize call for uuid map of referenceable nodes

michaelzangerle:
Fixes https://github.com/jackalope/jackalope-doctrine-dbal/issues/450 for 1.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-1.x
-===
+1.13.1
+------
+
+* Fixed cache key sanitize for UUID map of referenceable nodes.
 
 1.13.0
 ------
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-1.x
-===
```
dbu, line 5: this needs to be kept. i will merge on the cli and revert the removal of this line.

dbu (approved):
thanks for backporting https://github.com/jackalope/jackalope-doctrine-dbal/commit/8cddf1fe23bec02578845ab03b72f6451ebc34ff

we really should have made factory methods for those keys, rather than repeat how the key is built in multiple places... i am waiting for the CI to confirm there are no surprises and then will merge and tag the release.

dbu:
ah, those builds won't happen because they are on legacy ubuntu. i fixed the 1.x branch to run on latest. continuing this in #456

